PR #27, opened 2020-05-11
Issue 22

### LINKED ISSUE #22: -f flag can not handle an absolute path
**Describe the bug**
-f flag can not handle an absolute path. Since it appends `./` to the path, the path becomes invalid

**To Reproduce**
1. Pass the `-f` flag a file path that starts with `/`
2. See error

**Expected behavior**
If the passed path starts with a `/` the app should not prefix with `./`

**Screenshots**
image

drazisil:
closes #22 

Add detection for absolute paths and do not prefix them with `./`

Add dry-run tests for same.

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -117,18 +117,18 @@ function getAllFiles(projectRoot, dirPath, arrayOfFiles) {
 
   files.forEach(function(file) {
     if (
-      fs.statSync(dirPath + "/" + file).isDirectory() &&
+      fs.statSync(path.join(dirPath, file)).isDirectory() &&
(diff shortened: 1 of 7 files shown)


src/helpers/files.js
```diff
@@ -169,5 +182,6 @@ module.exports = {
   fetchGitRoot,
   parseGitIgnore,
   getCoverageFiles,
-  coverageFilePatterns
+  coverageFilePatterns,
+  getFilePath
```
thomasrockhu, line 186: super nit: if possible, if things can be alphabetized, I like them alphabetized

drazisil: PRs welcome ;)

PR #28, opened 2020-05-14
Add Alpine Linux support

drazisil:
Current plan is to create an APK.


drazisil:
Assign to @drazisil-codecov instead.

PR #49, opened 2021-04-13
chore: replace eslint with standard and lint

drazisil-codecov:


test/helpers/web.test.js
```diff
@@ -1,106 +1,102 @@
...
 
-describe("Web Helpers", function() {
-  let inputs;
-  let uploadURL;
-  let token;
-  let uploadFile;
-  let version;
-  let query;
-  beforeEach(function() {
-    inputs = { args: {}, envs: {} };
```
thomasrockhu, line 13: my node isn't great, I'm guessing this doesn't need to be declared here anymore?

thomasrockhu (approved):
Just a small comment

test/helpers/web.test.js
drazisil-codecov: Which line? (keeping mind this is a linting PR)

PR #75, opened 2021-05-04
Custom defined name of the upload

eddiemoore:
Adding name flag as defined in the upload API https://docs.codecov.io/reference#upload

Removed `expect` as it is not needed as jest has it built in.

Added JSDoc comments to some functions.
Adding coverage to index.js file

### DIFF
--- a/bin/codecov
+++ b/bin/codecov
@@ -26,6 +26,11 @@ var argv = require("yargs") // eslint-disable-line
       default: "",
       description: "Flag the upload to group coverage metrics"
     },
+    name: {
+      alias: "n",
+      default: "",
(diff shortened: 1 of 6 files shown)


drazisil-codecov (approved):
Looks great! Regarding the JSDoc comments, do you want to also add a jsconfig.js so VS Code can pick up on them?

eddiemoore:
> Looks great! Regarding the JSDoc comments, do you want to also add a jsconfig.js so VS Code can pick up on them?

@drazisil-codecov VSCode picks up on them without the jsconfig.js. It uses the inbuilt typescript engine to handle the comments

PR #76, opened 2021-05-05
Allow Environment variables

eddiemoore:
Specify environment variables to be included with this build.
Also accepting environment variables: `CODECOV_ENV`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
   build-linux-and-osx:
     docker:
       # specify the version you desire here
-      - image: circleci/node:12.16.1
+      - image: circleci/node:14.16.1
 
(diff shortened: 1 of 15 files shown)


ajbrown:
Please hold off on merging this in.  We are assessing whether this is needed any longer, and may remove it if not needed.

drazisil-codecov (changes requested):
@eddiemoore Can you have the test runs use this flag to pass an ENV? I want to ensure that Codecov won't crash if it's sent a report with one of these.

eddiemoore:
@drazisil-codecov yeah sure. I'll add that

drazisil-codecov (approved):
Good to merge once it passes, as long as Codecov can display it.

Issue #85, opened 2021-05-07
Does not scan for files on OSX or Windows


drazisil-codecov:
**Describe the bug**

While the uploader can correctly locate files without a `.git` directory, it appears unable to on Windows and OSX

**To Reproduce**
1. Run the uploader in a directory that is not a git repo, or a child of one.

**Expected behavior**

Should be able to locate report files, similar to on Linux https://app.circleci.com/pipelines/github/codecov/uploader/445/workflows/27d617a9-1375-4fdb-b077-39922becdf57/jobs/2686

**Screenshots**

**Additional context**

PR #92, opened 2021-05-12
fix no git repo error

eddiemoore:
Adding encoding to spawnSync to return as a string so it wont throw an error

Should help #85

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
       "it",
       "describe",
       "beforeEach",
-      "afterEach"
+      "afterEach",
+      "jest"
     ]
   }
 }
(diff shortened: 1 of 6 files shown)


drazisil-codecov (dismissed):
Verify that windows-without-git and osx-without-git jobs can upload to Codecov correctly before merging

drazisil-codecov (changes requested):
https://codecov.io/gh/codecov/uploader/commit/eab20eb6cb8695e67dbac960a09a2bc0eca6b9a4/build

Windows report is failing to process due to the windows uploader uploading the codecov.exe binary as a report file.

drazisil-codecov (approved):

PR #96, opened 2021-05-18
Add verbose logging

drazisil-codecov:
Currently , the `-v` vrbose flag does not do anything. Creating a logger wrapper and the tests, then will convert all output to use the logger.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
     ],
     "coverageReporters": [
       "text",
-      "cobertura"
+      "cobertura",
+      "html"
     ],
     "setupFilesAfterEnv": [
       "<rootDir>/test/test_helpers.js"
--- a/package.json
+++ b/package.json
@@ -60,8 +60,7 @@
     ],
     "coverageReporters": [
       "text",
-      "cobertura",
-      "html"
+      "cobertura"
     ],
     "setupFilesAfterEnv": [
       "<rootDir>/test/test_helpers.js"
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -122,7 +122,7 @@ function globBlacklist () {
     'remapInstanbul.coverage*.json',
     'scoverage.measurements.*',
     'test_*_coverage.txt',
-    'testrunner-coverage*',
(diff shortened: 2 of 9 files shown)


src/helpers/logger.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {string} options.level
+ * @param {Object} options.args
+ * @returns void
+ */
+function log (message, options) {
+  if (!options || !options.level) {
+    return console.log(message)
+  }
+  switch (options.level) {
+    case 'debug':
```
thomasrockhu, line 16: would prefer these to be constants

package.json
```diff
@@ -60,7 +60,8 @@
     ],
     "coverageReporters": [
       "text",
-      "cobertura"
+      "cobertura",
+      "html"
```
thomasrockhu, line 64: is this needed?

src/helpers/files.js
```diff
@@ -224,28 +225,30 @@ function parseGitIgnore (projectRoot) {
...
-        arrayOfFiles.push(
+      arrayOfFiles.push(
           `${path.join(dirPath.replace(projectRoot, '.'), file)}\n`
-        )
-      }
+      )
     }
   })
-
+  log(`Search complete for files in ${dirPath}`, { level: 'debug', args })
```
thomasrockhu, line 251: nit: not sure we need to specify again if we have the `dirPath` at the beginning of this method. doesn't matter ot me either way

drazisil-codecov: Because it's a nested recursive method.

package.json
drazisil-codecov: For my local sanity. Not really for production

src/helpers/logger.js
drazisil-codecov: As would I. I don't know how to do enums in JS though.

drazisil-codecov: I have made an attempt. Can you please take a look, @thomasrockhu ?

thomasrockhu (approved):

PR #103, opened 2021-05-27
Move discovered coverage reports to the trash

eddiemoore:
`-c` clean up coverage reports


drazisil-codecov (approved):
Coverage drop is ok, method is tested.

PR #116, opened 2021-06-03
test: Add CircleCI tests and refactor testing

thomasrockhu:
Refactors tests in `index.ts` to check for parameters
Adds tests for CircleCI provider
Removes the `private` function calls from API

### DIFF
--- a/src/ci_providers/provider_circleci.js
+++ b/src/ci_providers/provider_circleci.js
@@ -24,12 +24,7 @@ function _getBranch (inputs) {
 
 function _getSHA (inputs) {
   const { args, envs } = inputs
-  try {
-    const sha = envs.CIRCLE_SHA1
-    return args.sha || sha
-  } catch (error) {
-    throw new Error(`There was an error getting the commit SHA: ${error}`)
-  }
+  return args.sha || envs.CIRCLE_SHA1
 }
 
 function _getSlug (inputs) {
@@ -38,7 +33,7 @@ function _getSlug (inputs) {
   if (envs.CIRCLE_PROJECT_REPONAME !== '') {
     slug = `${envs.CIRCLE_PROJECT_USERNAME}/${envs.CIRCLE_PROJECT_REPONAME}`
   } else {
-    slug = `${envs.CIRCLE_REPOSITORY_URL}.git`
+    slug = `${envs.CIRCLE_REPOSITORY_URL.split(":")[1].split('.git')[0]}`
   }
   return args.slug || slug
 }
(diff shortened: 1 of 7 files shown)


src/ci_providers/provider_circleci.js
```diff
@@ -24,12 +24,7 @@ function _getBranch (inputs) {
 
 function _getSHA (inputs) {
   const { args, envs } = inputs
-  try {
-    const sha = envs.CIRCLE_SHA1
-    return args.sha || sha
-  } catch (error) {
-    throw new Error(`There was an error getting the commit SHA: ${error}`)
-  }
+  return args.sha || envs.CIRCLE_SHA1
```
drazisil-codecov, line 27: What happens in the case both are empty?

src/ci_providers/provider_circleci.js
```diff
@@ -38,7 +33,7 @@ function _getSlug (inputs) {
   if (envs.CIRCLE_PROJECT_REPONAME !== '') {
     slug = `${envs.CIRCLE_PROJECT_USERNAME}/${envs.CIRCLE_PROJECT_REPONAME}`
   } else {
-    slug = `${envs.CIRCLE_REPOSITORY_URL}.git`
+    slug = `${envs.CIRCLE_REPOSITORY_URL.split(":")[1].split('.git')[0]}`
```
drazisil-codecov, line 36: Will this handle https:// URLs as well?

test/providers/index.test.js
```diff
@@ -99,7 +65,18 @@ describe('CI Providers', () => {
             'remote.origin.url'])).thenReturn({
             stdout: 'http://github.com/testOrg/testRepo.git'
```
drazisil-codecov, line 66: This is what I mean by an https:// URL. It has no colon

drazisil-codecov (approved):
Approved, provided none of my comments are an issue.

src/ci_providers/provider_circleci.js
thomasrockhu: probably not, but this is how it's handled in the bash uploader currently

src/ci_providers/provider_circleci.js
thomasrockhu: has not been an issue on the bash uploader, assuming it's a standard env variable. in any case, it will return `undefined`

test/providers/index.test.js
thomasrockhu: I don't see a world where if this would have reared its head, it has been lying in wait all this time

https://github.com/codecov/codecov-bash/blob/master/codecov#L667

PR #134, opened 2021-06-11
test: exclude codecov.exe on windows

drazisil-codecov:
This fixes an issue where codecov.exe would attempt to upload itself on Windows, and the report processing would fail due to it being a binary.

### DIFF
--- a/test/dummy.codecov.exe
+++ b/test/dummy.codecov.exe
@@ -1,0 +1,1 @@
+I'm not really an exe ;)
--- a/test/helpers/files.test.js
+++ b/test/helpers/files.test.js
@@ -75,6 +75,10 @@ describe('File Helpers', () => {
         'dummy.codecov.exe'
       )
 
+      expect(results).not.toContain(
+        'codecov.exe'
+      )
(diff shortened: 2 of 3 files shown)


test/dummy.codecov.exe
```diff
@@ -0,0 +1 @@
+I'm not really an exe ;)
```
thomasrockhu, line 1: lol

test/helpers/files.test.js
```diff
@@ -4,6 +4,7 @@ const td = require('testdouble')
 const fs = require('fs')
 const childProcess = require('child_process')
 const fileHelpers = require('../../src/helpers/files')
+const { notDeepEqual } = require('assert')
```
thomasrockhu, line 7: I'd prefer this to be above the `child_process` require.

thomasrockhu: Also, this isn't used?

thomasrockhu (dismissed):
Small nits

thomasrockhu (approved):

PR #141, opened 2021-06-15
fix: use posix path

drazisil:
This changes `path` to use `path.posix` per https://nodejs.org/dist/latest-v14.x/docs/api/path.html#path_windows_vs_posix

Thanks @eddiemoore !

This also runs tests on Windows prior to building.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,9 +67,9 @@ jobs:
     steps:
       - checkout
       - attach_workspace:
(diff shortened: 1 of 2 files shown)


drazisil:
Looks great! https://codecov.io/gh/codecov/uploader/tree/0f6f6018eb13262761ae0a3c8d40b4b828b562a7

As far as the coverage drop... it doesn't like blank lines? 🤷‍♀️

hootener (approved):

PR #143, opened 2021-06-15
fix: paths work correctly on windows

drazisil-codecov:


thomasrockhu (approved):

RA80533:
Where is the fix? It appears this PR only changes the README.md.

RA80533:
It looks like 186f731 contains a fix related to POSIX paths instead of Windows.

Was there a previous issue with paths not working correctly in one of the two OS environments?

PR #147, opened 2021-06-18
fix: correct flags regex to match server and docs

### LINKED ISSUE #146: Flags not working when they contain a period
**Describe the bug**

The `-F` command line flag is not working correctly.  Flag used was `12.x.v2`

**To Reproduce**
* Attempt to upload using the `-F` flags option

**Expected behavior**

Flags are shown in the UI and the PR comment

https://codecov.io/gh/drazisil/sigdb/commit/ddfcff46193dd3c19ef15d3673d4c27840beb18a/build

image

https://github.com/drazisil/sigdb/pull/1/commits/ddfcff46193dd3c19ef15d3673d4c27840beb18a

drazisil:
fixes #146

### DIFF
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^(?!-\.)[\w,-]+$/
+  const mask = /^(?!-.)[\w,\.-]+$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w\.\-]+${1,45}/
+  const mask = /^[\w\.\-]+{1,45}$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[a-z0-9_\.\-]{1,45}$/
+  const mask = /^[\w\.\-]+${1,45}/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w,\.,\-]{1,45}$/
+  const mask = /^[\w\.\-]{1,45}$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w\.\-]+{1,45}$/
+  const mask = /^[\w,\.,\-]+{1,45}$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^(?!-.)[\w,\.-]+$/
+  const mask = /^[a-z0-9_\.\-]{1,45}$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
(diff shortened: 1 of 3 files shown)


drazisil:
I disagree that https://app.codecov.io/gh/codecov/uploader/compare/147/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=codecov#diff-c3JjL2NpX3Byb3ZpZGVycy9wcm92aWRlcl9jaXJjbGVjaS5qcw== is a coverage drop. 🤷‍♀️

RA80533:
What is a flag? What should it be validated for? It seems like this is early to validate for anything other than that flags are of type `string`.

It might be best to use a command-line parsing library for this functionality. Even with the addition of periods to the regular expression, validation would fail on other seemingly “valid” (from the perspective of them being genuine) flags.

drazisil:
> What is a flag? What should it be validated for? It seems like this is early to validate for anything other than that flags are of type `string`.
> 
> It might be best to use a command-line parsing library for this functionality. Even with the addition of periods to the regular expression, validation would fail on other seemingly “valid” (from the perspective of them being genuine) flags.

A "flag" in this context is a feature of Codecov https://docs.codecov.com/docs/flags

It's being validated because not all strings are accepted and it's preferred to stop at the upload, rather then at a server error.

RA80533:
> Only lowercase, alphanumeric values are accepted. ^[a-z0-9_\.\-]{1,45}$

Oh, I see. Thanks for the link.

drazisil:
> Should the regex be changed to the snippet from the documentation?

Indeed. Turns out the docs were incorrect as well. Now they both match the server :)

src/helpers/validate.js
```diff
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^(?!-)[\w,-]+$/
+  const mask = /^[\w,\.,\-]+{1,45}$/
```
drazisil, line 12: `const mask = /^[\w,\.,\-]{1,45}$/`

drazisil:
Ok, @codecov/open-source . This is ready for review and merge. Finally.

src/helpers/validate.js
```diff
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
```
RA80533, line 11: Should this be renamed to `validateFlag` (singular) since it validates *one* flag?

src/helpers/validate.js
RA80533: There is benefit in moving this regex outside.

Currently, every invocation of `validateFlags` will instantiate a unique regular expression instance which will promptly be destroyed (and never reused) after the function returns. The optimization here is to simply instantiate one regular expression but share it among every invocation of the function.

src/helpers/validate.js
```diff
@@ -9,7 +9,7 @@ function validateURL(url) {
...
-  const mask = /^(?!-)[\w,-]+$/
+  const mask = /^[\w,\.,\-]{1,45}$/
```
RA80533, line 12: There is benefit in moving this regex outside.

Currently, every invocation of `validateFlags` will instantiate a unique regular expression instance which will promptly be destroyed (and never reused) after the function returns. The optimization here is to simply instantiate one regular expression but share it among every invocation of the function. This is common practice with regular expressions in JavaScript because complex patterns can be (relatively) difficult to compile.

RA80533: Is this intended to match commas?

RA80533:
The boundaries of the regular expression should be tested; namely:
- Length: (0, 1, 45, 46)
- Characters: (uppercase and lowercase alphabetical; numeral; underline; dash; period; comma)

RA80533:
> > Should the regex be changed to the snippet from the documentation?
> 
> Indeed. Turns out the docs were incorrect as well. Now they both match the server :)

It appears that the docs are still somewhat incorrect:

> `^[\w\.\-]+${1,45}$`

`+$` should be removed, resulting in:

> `^[\w\.\-]{1,45}$`

src/helpers/validate.js
drazisil-codecov:
```suggestion
  const mask = /^[\w\.\-]{1,45}$/
```
fix: do not match commas

src/helpers/validate.js
drazisil-codecov: Probably , but that's scope creep imo.

src/helpers/validate.js
drazisil-codecov: > Currently, every invocation of validateFlags will instantiate a unique regular expression instance which will promptly be destroyed (and never reused) after the function returns

I don't follow. Except in tests, this regex should generally only be used once per execution

PR #151, opened 2021-06-21
Optimize glob functionality

### LINKED ISSUE #154: index.js#L116
https://github.com/codecov/uploader/blob/821fbb5c894818ce6a0c9118744a5957c7ecf675/src/index.js#L116

RA80533:
Fixes #154

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -55,7 +55,7 @@ describe('Uploader Core', () => {
     process.env.SOMETHING = 'red'
     process.env.ANOTHER = 'blue'
     const log = jest.spyOn(console, 'log').mockImplementation(() => {})
-    jest.spyOn(process, 'exit').mockImplementation(() => {})
+    jest.spyOn(process, 'exit').mockImplementationOnce(() => {})
     await app.main({
       name: 'customname',
       token: 'abcdefg',
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -7,10 +7,6 @@ const fs = require('fs')
 describe('Uploader Core', () => {
   const env = process.env
   
-  beforeEach(() => {
-    jest.spyOn(process, 'exit').mockImplementation(() => {})
-  })
-
   afterEach(() => {
     process.env = env
     jest.restoreAllMocks()
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -55,7 +55,7 @@ describe('Uploader Core', () => {
     process.env.SOMETHING = 'red'
     process.env.ANOTHER = 'blue'
     const log = jest.spyOn(console, 'log').mockImplementation(() => {})
-    const log = jest.spyOn(process, 'exit').mockImplementation(() => {})
+    jest.spyOn(process, 'exit').mockImplementation(() => {})
     await app.main({
       name: 'customname',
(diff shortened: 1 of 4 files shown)


RA80533:
The linter split a few too many expressions into multiple lines.

RA80533:
Any idea what failed?

RA80533:
I believe this is due to how process.exit needs to be mocked. There are a few tests in index.test.js that don't mock it but the invocations to `main()` nonetheless call process.exit.

test/index.test.js
```diff
@@ -54,7 +54,6 @@ describe('Uploader Core', () => {
   it('Can parse environment variables', async () => {
     process.env.SOMETHING = 'red'
     process.env.ANOTHER = 'blue'
-    jest.spyOn(process, 'exit').mockImplementation(() => {})
     const log = jest.spyOn(console, 'log').mockImplementation(() => {})
     await app.main({
```
drazisil-codecov, line 58: `test/index.test.js:58:5`

https://app.circleci.com/pipelines/github/codecov/uploader/678/workflows/96eeefce-0f14-4972-988e-2b3dd3039c7f/jobs/4549?invite=true#step-105-43

Ok, so it _has_ to be mocked. 👎

drazisil-codecov:
```suggestion
     const log = jest.spyOn(process, 'exit').mockImplementation(() => {})
    await app.main({
```

drazisil-codecov: Going to try mocking just this one, to see if the error moves.

test/index.test.js
```diff
@@ -54,8 +54,8 @@ describe('Uploader Core', () => {
   it('Can parse environment variables', async () => {
     process.env.SOMETHING = 'red'
     process.env.ANOTHER = 'blue'
-    jest.spyOn(process, 'exit').mockImplementation(() => {})
     const log = jest.spyOn(console, 'log').mockImplementation(() => {})
+    const log = jest.spyOn(process, 'exit').mockImplementation(() => {})
```
drazisil-codecov, line 58:
```suggestion
    jest.spyOn(process, 'exit').mockImplementation(() => {})
```
Do no try to code before caffeine 🤦

RA80533:
The use of process.exit here is somewhat unconventional. Typically, errors are thrown in place of where process.exit is used such that the handling function can do things like:
- Print the error to the user
- Log any diagnostics related to the error
- Decide whether to exit and, if so, the status code with which to exit (set via `process.exitCode` rather than `process.exit()` as recommended by the official Node.js documentation)

RA80533:
That pattern should only require a slight refactor. Binary-like executables are at a low enough level where handling exit codes are very often something the developer wishes to have explicit control over so it's understandable why it was written in such a manner.

Most of the functionality related to handling the way the process exits can be moved over to the file in the "bin" folder. One benefit of doing error handling this way is that the errors being thrown can be tested through Jest's built-in matchers: `toThrow()`, `toThrowErrorMatchingSnapshot()`, etc.

RA80533:
No worries. My contributions are for the project, even if those contributions don't come in the form of literal changes to the codebase. Ideas are contributions too.

PR #173, opened 2021-07-02
Add vendor dir to file blacklist

montross50:
This should allow php projects to have reasonable performance when using the uploader.

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -24,6 +24,7 @@ function manualBlacklist() {
     '.nvmrc',
     '.gitignore',
     '.DS_Store',
+    'vendor',
   ]
 }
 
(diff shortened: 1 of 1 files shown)


src/helpers/files.js
```diff
@@ -24,13 +24,15 @@ function manualBlacklist() {
...
     '.gitignore',
     '.DS_Store',
+    'vendor',
   ]
 }
 
 function globBlacklist() {
   // TODO: honor the .gitignore file instead of a hard-coded list
   return [
     'node_modules/**/*',
```
drazisil-codecov, line 34: @drazisil-codecov , why do you do the one directory this way?

drazisil-codecov (approved):

src/helpers/files.js
thomasrockhu: lolol

montross50: I was just copying what was there for other dirs. Not sure the best approach here.

drazisil-codecov: I'm the last person to ask about globs. In theory my tests says both ways work 😇 🤷‍♀️

PR #179, opened 2021-07-06
Adding buildkite provider

eddiemoore:


src/ci_providers/provider_buildkite.js
```diff
@@ -0,0 +1,126 @@
+// https://buildkite.com/docs/guides/environment-variables
```
thomasrockhu, line 1: can we remove this comment ish?

eddiemoore: Yeah sure
```suggestion
```

thomasrockhu: Sorry, I meant all of it for all the methods as in the other providers that aren't the template

drazisil-codecov: I actually like them, so as when they change, we know where to look to see what they changed to. 🤷‍♀️

drazisil-codecov (approved):

PR #181, opened 2021-07-06
Adding drone provider

eddiemoore:


src/ci_providers/provider_drone.js
```diff
@@ -0,0 +1,65 @@
+function detect(envs) {
+  return !!envs.DRONE
```
drazisil-codecov, line 2: I'm not familiar with the `!!` pattern, what does this do?

drazisil-codecov (approved):

src/ci_providers/provider_drone.js
thomasrockhu: returns a boolean as opposed to possibly not a boolean

drazisil-codecov: TIL 🎉

eddiemoore: yeah means if the value is "truthy" it will return `true`

RA80533: It's a more terse way of checking for truthiness compared to using something like the equality operator. Two common points against its use I commonly see are that it:

1. **Obscures intent for those unfamiliar with its syntax**
   
2. **Can introduce hard-to-find errors**
   An extra `(` will introduce a syntax error that will be immediately apparent to the developer. An extra `!` will not.

The `Boolean()` function (not to be confused with the `new`-able Boolean constructor) is likely a more understandable way to convey the purpose of checking for truthiness for some people although it's certainly rooted in opinion.

PR #182, opened 2021-07-06
refactor: update to use spawnsync

drazisil:
* refactor: change exec Sync to use spawnSync. I believe this was part of the fix for Command Injection on codecov-node, can't confirm. @eddiemoore and @RA80533, would love your review.
* test: add test and alert for execSync use to test/test_helpers.js

### DIFF
--- a/src/ci_providers/provider_azurepipelines.js
+++ b/src/ci_providers/provider_azurepipelines.js
@@ -57,9 +57,10 @@ function _getSHA(inputs) {
 
   if (_getPR(inputs)) {
     const mergeCommitRegex = /^[a-z0-9]{40} [a-z0-9]{40}$/
-    const mergeCommitMessage = childProcess.execSync(
-      'git show --no-patch --format="%P"',
-    )
+    const mergeCommitMessage = childProcess
+      .spawnSync('git', ['show', '--no-patch', '--format="%P"'])
+      .stdout.toString()
+      .trimRight()
     if (mergeCommitRegex.exec(mergeCommitMessage)) {
       const mergeCommit = mergeCommitMessage.split(' ')[1]
       log(`    Fixing merge commit SHA ${commit} -> ${mergeCommit}`)
--- a/src/ci_providers/provider_azurepipelines.js
+++ b/src/ci_providers/provider_azurepipelines.js
@@ -58,8 +58,8 @@ function _getSHA(inputs) {
   if (_getPR(inputs)) {
     const mergeCommitRegex = /^[a-z0-9]{40} [a-z0-9]{40}$/
     const mergeCommitMessage = childProcess
-      .spawnSync('git', ['show', '--no-patch', '--format="%P"'])
-      .stdout.toString()
+      .execFileSync('git', ['show', '--no-patch', '--format="%P"'])
+      .toString()
       .trimRight()
     if (mergeCommitRegex.exec(mergeCommitMessage)) {
       const mergeCommit = mergeCommitMessage.split(' ')[1]
--- a/src/helpers/git.js
+++ b/src/helpers/git.js
@@ -24,9 +24,21 @@ function parseSlug(slug) {
 function parseSlugFromRemoteAddr(remoteAddr) {
   let slug = ''
   if (!remoteAddr) {
-    remoteAddr = childProcess.execSync(
-      "git config --get remote.origin.url || hg paths default || echo ''",
-    )
+    remoteAddr = childProcess
+      .spawnSync('git', [
+        'config',
+        '--get',
+        'remote.origin.url',
+        '||',
+        'hg',
+        'paths',
+        'default',
+        '||',
+        'echo',
+        "''",
+      ])
+      .stdout.toString()
+      .trimRight()
   }
   if (remoteAddr) {
     slug = parseSlug(remoteAddr)
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w\.\-]{1,45}$/
+  const mask = /^[\w.-]{1,45}$/
   return mask.test(flags)
 }
 
--- a/src/helpers/validate.js
+++ b/src/helpers/validate.js
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w.-]{1,45}$/
+  const mask = /^[\w\.\-]{1,45}$/
   return mask.test(flags)
 }
 
--- a/test/providers/provider_azurepipelines.test.js
+++ b/test/providers/provider_azurepipelines.test.js
@@ -140,10 +140,13 @@ describe('Jenkins CI Params', () => {
       service: 'azure_pipelines',
       slug: '',
     }
-    const execSync = td.replace(childProcess, 'execSync')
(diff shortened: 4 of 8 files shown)


RA80533:
spawnSync is in the right direction (the docs are [here](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options)). From the perspective of command injection, there's one big difference between spawnSync and execSync that sets spawnSync apart from execSync: spawnSync does not open a shell to execute commands (unless it is configured to so).

This is easier to understand the implications of when making the comparison between execSync and execSyncFile. If execSyncFile is passed `"node"` as its file/command, for example, it will directly interface with the Git binary. Bash (or any other shell, for that matter) will never be used which means things like input redirection will simply not be available (which is a good thing from a security perspective).

- 👎 execSync
- 👍 execFileSync, spawnSync

drazisil:
Thanks, @RA80533 ! That was what I thought I remembered, but I'm "out sick" today due to a lapse in depression med, so I couldn't/can't currently grok that out of the docs. Glad to hear my memory still works. 😜

drazisil:
> `execFileSync(…)` could be replaced with `await execFile(…)` to minimize idling.

Does that work nativity without needing to promisfy it?

RA80533:
> but I'm "out sick" today due to a lapse in depression med

I'm awfully sorry to hear that. Lapses in medication are not fun at all. I hope the feedback from the medication isn't too long and you're feeling better soon!

RA80533:
> Does that work nativity without needing to promisfy it?

It's still not native so it would have to be wrapped with util.promisify. It would only be necessary to do this once as long as its exported and used everywhere else.

src/ci_providers/provider_azurepipelines.js
```diff
@@ -57,9 +57,10 @@ function _getSHA(inputs) {
 
   if (_getPR(inputs)) {
     const mergeCommitRegex = /^[a-z0-9]{40} [a-z0-9]{40}$/
-    const mergeCommitMessage = childProcess.execSync(
-      'git show --no-patch --format="%P"',
-    )
+    const mergeCommitMessage = childProcess
+      .spawnSync('git', ['show', '--no-patch', '--format="%P"'])
+      .stdout.toString()
```
RA80533, line 62:
```suggestion
      .execFileSync('git', ['show', '--no-patch', '--format="%P"'])
      .toString()
```

:blush:

test/test_helpers.js
```diff
@@ -1,4 +1,29 @@
...
+beforeAll(() => {
+  execSync = jest.spyOn(childProcess, 'execSync').mockImplementation(() => {
+    throw new Error(
+      `Security alert! Do not use execSync(), use spawnSync() instead`,
+    )
+  })
+  exec = jest.spyOn(childProcess, 'exec').mockImplementation(() => {
+    throw new Error(`Security alert! Do not use exec(), use spawn() instead`)
+  })
+})
```
RA80533, line 20: This is super neat.

src/helpers/validate.js
```diff
@@ -9,7 +9,7 @@ function validateURL(url) {
 }
 
 function validateFlags(flags) {
-  const mask = /^[\w\.\-]{1,45}$/
+  const mask = /^[\w.-]{1,45}$/
```
RA80533, line 12: This might be better to leave it as is because it shows that *those are definitely* the literal characters. An unescaped period might appear to be a wildcard even though it actually matches a literal period.

src/helpers/git.js
```diff
@@ -24,9 +24,21 @@ function parseSlug(slug) {
...
+        '--get',
+        'remote.origin.url',
+        '||',
+        'hg',
+        'paths',
+        'default',
+        '||',
+        'echo',
+        "''",
+      ])
```
RA80533, line 39: This will have to be split up into two separate calls to spawnSync:
- Try to call `git`
- Fallback to `hg`

Lines 28 to 38 are arguments passed to Git. If spawnSync was given `'bash'` instead of `'git'` and `'git'` was placed at the start of the array, this kind of pattern would work but it would be the equivalent of calling execSync.

sbeidas (approved):

sbeidas:
this looks great thank you for contributing @RA80533  🙏

src/helpers/validate.js
drazisil: The linter yelled at me for that one. 😆

src/helpers/git.js
drazisil: @thomasrockhu do we have any metrics on what percentage of our customers use mercurial? I'm very tempted to drop support for it here.

src/ci_providers/provider_azurepipelines.js
drazisil: @ra80533 this appears to have caused a failure in tests. It's because we didn't update tests, right?

RA80533: If the tests are hooking up into spawnSync, yes.

test/providers/provider_azurepipelines.test.js
```diff
@@ -140,10 +140,13 @@ describe('Jenkins CI Params', () => {
...
-    td.when(execSync('git show --no-patch --format="%P"')).thenReturn(
-      'testingsha123456789012345678901234567890 testingmergecommitsha2345678901234567890',
-    )
+    const spawnSync = td.replace(childProcess, 'spawnSync')
+    td.when(
+      spawnSync('git', ['show', '--no-patch', '--format="%P"']),
+    ).thenReturn({
+      stdout:
+        'testingsha123456789012345678901234567890 testingmergecommitsha2345678901234567890',
+    })
```
RA80533, line 149:
```suggestion
    const execFileSync = td.replace(childProcess, 'execFileSync')
    td.when(
      execFileSync('git', ['show', '--no-patch', '--format="%P"']),
    ).thenReturn('testingsha123456789012345678901234567890 testingmergecommitsha2345678901234567890')
```

src/ci_providers/provider_azurepipelines.js
drazisil: Yah, every thing is mocked, as the results of git in ci are non-deterministic

test/providers/provider_azurepipelines.test.js
RA80533: I believe this is the pattern that would allow the tests to function correctly.

src/helpers/validate.js
drazisil: Do you mind suggesting it back? I'm afk on mobile, so can't make the code changes myself right now.

src/helpers/git.js
drazisil: > EDIT: I hope that paints a better picture of how spawnSync and execFileSync differ from execSync.

It does, thanks. I was suspicious that would fail, I've never encountered mercurial so was not sure how to test.

test/providers/provider_azurepipelines.test.js
drazisil: 🤞 I think so too

src/helpers/validate.js
RA80533, line 12:
```suggestion
  const mask = /^[\w\.\-]{1,45}$/
```

src/helpers/git.js
thomasrockhu: unfortunately we don't @drazisil, it would be worth adding that metadata

drazisil-codecov: Created https://github.com/codecov/uploader/issues/183

PR #186, opened 2021-07-07
feat: add bitbucket pipelines support

drazisil-codecov:
* feat: add bitbucket pipelines support
* feat: add helper method to validate


src/helpers/validate.js
```diff
@@ -18,9 +19,22 @@ function validateFileNamePath(path) {
   return mask.test(path)
 }
 
+/**
+ * Validate that a SHA is the correct length and content
+ * @param {string} commitSHA
+ * @param {number} requestedLength
+ * @returns {boolean}
+ */
+function validateSHA(commitSHA, requestedLength) {
```
thomasrockhu, line 28: my JS is trash, is there a way to mark `requestedLength` as optional?

thomasrockhu (approved):
barring optional changes

src/helpers/validate.js
drazisil-codecov: Yes. What do you want for the default? 40?

src/ci_providers/provider_bitbucket.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+function _getSlug(inputs) {
+  const { args, envs } = inputs
+
+  let slug = ''
+  if (envs.BITBUCKET_REPO_OWNER && envs.BITBUCKET_REPO_SLUG) {
+    slug = `${envs.BITBUCKET_REPO_OWNER}/${envs.BITBUCKET_REPO_SLUG}`
+  }
+  return args.slug || slug
```
RA80533, line 58: Be aware that `slug` is sourced from the environment. This is a not bad thing by any means but it's important to make sure this is never provided to something that expects a sanitized string.

Since this project is written in JavaScript rather than TypeScript, tracking down these sources of tainted data may prove to be quite difficult. CodeQL is marketed as doing this kind of tracking but it is probably not mature enough to be completely reliant upon.

It would be worth investigating TypeScript for use in this project—worth its weight in gold, in fact, because an effective TypeScript configuration would catch the most common classes of errors that happen in JavaScript.

src/helpers/validate.js
RA80533: > 40?

For reference, full-length Git SHAs are 40 characters long (40 hexidecimal characters which, translated to ASCII or UTF-8, etc., equates to 40 characters).

```suggestion
const GIT_SHA_LENGTH = 40;

function validateSHA(commitSHA, requestedLength = GIT_SHA_LENGTH) {
```

src/ci_providers/provider_bitbucket.js
drazisil-codecov: Indeed, and I'd greatly prefer that to JSDocs. I'd make some _mumble mumble_ about a start date of Nov 21, 2019 and a contributor % https://github.com/codecov/uploader/graphs/contributors, but that's me still making excuses for my less them perfect JS skills lol.

Good catch about the envs though, will validate them.

drazisil-codecov: @thomasrockhu what's a good regex mask for a slug? It always has to be `^[\w\d\-\/]+$`, correct?

drazisil-codecov: Ugg, no. I need support for international unicode chars as well. 🤔

drazisil-codecov: https://github.com/codecov/uploader/issues/187 for TypeScript

drazisil-codecov: I'm going to sidebar validating slug to https://github.com/codecov/uploader/issues/188, so I don't scope creep too much. The Bitbucket support is blocking a customer trialing it. Adding a note for myself to see if we are passing an unfiltered slug in the other providers.

RA80533: > Good catch about the envs though, will validate them.

With regards to validating things like slugs (or names, addresses, etc.), unless there is some definite shape to the piece of data (which should be guided by, at most, technical limitations) from which a validation algorithm can be defined, validation may not be necessary.

Often times, it is assumed that there is a need for validation simply on the basis of having a piece of data to validate. For example, what does a valid name look like in the context of an electronic form for a doctor's office? There's not really a definite answer because a person's name can be anything. There's [a great blog post from 2010](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) that touches upon this kind of area quite well (please excuse the crassness towards the end).

In the context of `slug`, *sanitization* would certainly applicable if the value is ever used in place of a string wherein certain characters must be escaped.

drazisil-codecov: Indeed , I'm mixing sanitize and validate, when they are not. As far as validate, there is a technical limit on some of these, and I'd rather it get stopped client side then at the server, though I'm open to discussion on that.

Regarding sanitize, That's similar. For example, how do I make sure that we don't block the customers who use emojis for their branch names? (yes, they do!)

RA80533: > I'd rather it get stopped client side then at the server

It's great that you brought this up.

In a client-server model, the server should be entirely authoritative of information it has ownership of. In terms of implementation, the server would validate everything sent to it by the client that would normally require validation. This is because the server does not have actual control over what the clients do and how information is generated. Server-side validation, however, does have the benefit of reusability across clients. Web, mobile, desktop, etc., can all send the same type of data to the same endpoint and receive the same responses.

The client *can* also perform validation (this is a common pattern) to provide quicker feedback to the user of any invalid fields. To alleviate the code duplication from this strategy, validation code can be modularized in a manner such that both the client and the server use the same functions, etc., although this is often not applicable if the client is written in a different language from to the server. This is actually a major reason why a lot of developers choose to run their code on Node.js.

I believe GitHub, for example, has most of its domain-specific input validation on the server but the web client is able to dynamically request validation of specific fields to show the user as they type (see [github/auto-check-element](https://github.com/github/auto-check-element)). It also uses native HTML input mechanisms to do basic input type validation for certain input fields although those are by no means authoritative.

.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
             npm ci
             npm test
             mkdir -p coverage-alpine
-            cp coverage/* coverage-alpine/
+            cp -r coverage/* coverage-alpine/
```
RA80533, line 103: Note that the `*` (glob) used on [line 103](https://github.com/codecov/uploader/pull/186/files/54e0992fcfc063877ae93c4e30858ef5d15cbd10..f6c38e0a43e9ea6c8224fcb2c0aa16ae651d0d3e#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R103) will expand to *only* include files that do *not* begin with a dot in Bash. In other words, if any files begin with a period, they will not be copied over to the destination.

I'm not sure how pertinent that information is but it may be relevant in the future if it's ever expected for a `.metadata` file or something to be copied.

RA80533, line 103:
```suggestion
            cp -r coverage/ coverage-alpine/
```

drazisil-codecov: I don't think it matters, but there is always the possibility of us switching the coverage reporter and wondering why our CI fails :)

src/ci_providers/provider_bitbucket.js
drazisil-codecov: > In a client-server model, the server should be entirely authoritative of information it has ownership of. 

I'm loving this discussion. Am I understanding that, unless it's a direct consumer of a piece of data, a method or system should not care if the data is valid? There's probably a better way to word that, but it makes sense.

On a similar topic, how much validation of it's inputs should a method do? I waver between "throw errors if it isn't good out of the gate" and "trust that the caller knows what I need". Of these two, the first is probably better. Typescript helps a lot here, but it doesn't cover runtime validation.

.circleci/config.yml
drazisil-codecov: That actually broke Alpine https://app.circleci.com/pipelines/github/codecov/uploader/738/workflows/09e0e139-b0c4-4492-b702-e90d4af53928/jobs/5134

Seems the copy didn;t find the coverage report with the change https://app.circleci.com/pipelines/github/codecov/uploader/738/workflows/09e0e139-b0c4-4492-b702-e90d4af53928/jobs/5130

src/ci_providers/provider_bitbucket.js
RA80533: > Am I understanding that, unless it's a direct consumer of a piece of data, a method or system should not care if the data is valid? There's probably a better way to word that, but it makes sense.

Essentially so.

There is tremendous strength in your understanding given the phrase *should not care*. This treads into the territory of architectural design but there is something called separation of concerns that broadly describes this approach.

> On a similar topic, how much validation of it's inputs should a method do? I waver between "throw errors if it isn't good out of the gate" and "trust that the caller knows what I need". Of these two, the first is probably better. Typescript helps a lot here, but it doesn't cover runtime validation.

Let me give an example of a short refactoring:

```js
const { readFile, stat } = require("fs/promises");

// (1)
async function read(path) {
  if (await stat(path)) {
    const contents = await readFile(path);
    
    // (A)
  } else {
    // (B)
  }
  
  // (C)
}

// (2)
async function read(path) {
  const contents = await readFile(path);
  
  // (A)
}
```

RA80533: - Access to the underlying system resources is requested via the `stat` and `readFile` functions
- They are handled by the OS and does not need to occur sequentially as the code would suggest
- Both can throw an `ENOENT: no such file or directory` error

.circleci/config.yml
RA80533: I am very curious as to why that might be. This is a good candidate for `git revert`.

PR #194, opened 2021-07-10
Adding AWS Codebuild provider

eddiemoore:


test/providers/provider_codebuild.test.js
```diff
@@ -0,0 +1,93 @@
...
+          CI: true,
+          CODEBUILD_CI: true,
+        },
+      }
+      const detected = providerCodeBuild.detect(inputs.envs)
+      expect(detected).toBeTruthy()
+    })
+  })
+
+  it('gets correct params', () => {
```
RA80533, line 33: Nest this under a `describe` block for `getServiceParams()` since that's the function being tested here.

src/ci_providers/provider_codebuild.js
```diff
@@ -0,0 +1,83 @@
...
+
+function _getSlug(inputs) {
+  const { args, envs } = inputs
+  return (
+    args.slug ||
+    (envs.CODEBUILD_SOURCE_REPO_URL
+      ? envs.CODEBUILD_SOURCE_REPO_URL.replace(/^.*github.com\//, '').replace(
+          /\.git$/,
+          '',
+        )
```
RA80533, line 61: This will only work for GitHub repositories. In other words, AWS repositories will not work.

eddiemoore: Yeah this was a direct copy from the node repo. We could change it to something else

RA80533: I *just* realized that. Since most of the functionality over there is being ported over here, are there explicit plans to deprecate that project?

eddiemoore: I think so.

drazisil-codecov: > This will only work for GitHub repositories. In other words, AWS repositories will not work.

AWS repositories are not supported by Codecov, we have no system on place to retrieve code from them.

We support GitHub, Bitbucket , and GitLab only.

test/providers/provider_codebuild.test.js
```diff
@@ -0,0 +1,96 @@
+const td = require('testdouble')
+const { describe } = require('yargs')
```
RA80533, line 2: This is the wrong kind of describe block.

RA80533: This was likely inserted automatically by your IDE. Test fixtures are typically injected into the environment as globals.

drazisil-codecov:
```suggestion
```

drazisil-codecov (changes requested):

drazisil-codecov (approved):

PR #195, opened 2021-07-10
Adding Wercker provider

eddiemoore:


src/ci_providers/provider_wercker.js
```diff
@@ -0,0 +1,66 @@
...
+
+function _getBranch(inputs) {
+  const { args, envs } = inputs
+
+  return args.branch || envs.WERCKER_GIT_BRANCH
+}
+
+// eslint-disable-next-line no-unused-vars
+function _getJob(envs) {
+  return ''
```
drazisil-codecov, line 23: This should take the `args` value of job

drazisil-codecov (changes requested):

src/ci_providers/provider_wercker.js
eddiemoore: We don't pass that in that at the moment.

drazisil-codecov: Huh. My bad, I thought we did. 🤦

drazisil-codecov (approved):

PR #196, opened 2021-07-10
Adding TeamCity provider

eddiemoore:


src/ci_providers/provider_teamcity.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function _getPR(inputs) {
+  const { args } = inputs
+  return args.pr || ''
+}
+
+// eslint-disable-next-line no-unused-vars
+function _getJob(envs) {
+  return ''
```
drazisil-codecov, line 47: This should take the job value from `args`

drazisil-codecov (changes requested):

src/ci_providers/provider_teamcity.js
eddiemoore: we don't pass that in at the moment

PR #200, opened 2021-07-13
fix: Allow multiple files to be read

thomasrockhu:
If multiple files are supplied to the uploader, it currently pushes all of them as the first element of an array e.g.

```
[
  [ first_file, second_file ]
]
```

This breaks the uploader's ability to read the files.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -105,6 +105,7 @@ async function main(args) {
   // Look for files
   let coverageFilePaths = []
   if (!args.file) {
+    log('Searching for coverage files...')
     coverageFilePaths = fileHelpers.getCoverageFiles(
       args.dir || projectRoot,
       // TODO: Determine why this is so slow (I suspect it's walking paths it should not)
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -119,9 +119,15 @@ async function main(args) {
...
+      for (const file of args.file) {
+        coverageFilePaths.push(validateHelpers.validateFileNamePath(file) ? file : '')
+      }
+    } else {
+      coverageFilePaths[0] = validateHelpers.validateFileNamePath(args.file)
+        ? args.file
+        : ''
+    }
     if (coverageFilePaths.length === 0) {
       throw new Error('Not coverage file found, exiting.')
```
drazisil-codecov, line 132:
```suggestion
      throw new Error('No coverage files found, exiting.')
```

drazisil-codecov (changes requested):
Is this covered by tests? Is there a way to add this to the dry runs in CI to make sure it's correctly handled?

drazisil-codecov (approved):

PR #203, opened 2021-07-15
feat: Allow for source from wrappers

thomasrockhu:
This will allow for more usage analytics on how the uploader is being used.


drazisil-codecov (approved):
Good catch

PR #212, opened 2021-07-16
fix: add token querystring

drazisil-codecov:


vlad-ko (approved):
👍

src/helpers/web.js
```diff
@@ -75,7 +75,7 @@ async function uploadToCodecovPUT(uploadURL, uploadFile) {
 async function uploadToCodecov(uploadURL, token, query, uploadFile, source) {
   try {
     const result = await superagent
-      .post(`${uploadURL}/upload/v4?package=${getPackage(source)}&${query}`)
+      .post(`${uploadURL}/upload/v4?package=${getPackage(source)}&token=${token}&${query}`)
```
RA80533, line 78: ☠️

RA80533:
The token is exposed as part of the destination for the HTTP/HTTPS request being sent. Since the destinations must be parsed in order to route them along to the destination itself, each node along the route a given request takes will see the token.

Issue #210, opened 2021-07-16
Prettier and lint-staged compete against each other


RA80533:
1. **Save a file targeted by Prettier**
   It will be modified to conform with the configured Prettier presets.
2. **Save the modifications made by Prettier**
   Prettier will run again but make no changes.
3. **Stage and commit the change(s) made by Prettier**
   lint-staged will modify the staged commits to conform with its own presets.

drazisil:
I feel like there should be a way to tell lint-staged to behave. @drazisil-codecov should investigate this.

drazisil-codecov:
@RA80533 I have not encountered this.

I believe this code handles your use case, can you confirm?
https://github.com/codecov/uploader/blob/master/package.json#L65-L69

mitchell-codecov:
I narrowed it down *that* in addition to the project-wide VS Code configuration in [.vscode/settings.json](https://github.com/codecov/uploader/blob/4ffeb72e8f7be7038813e4bdfe04fbf63ec4d0d3/.vscode/settings.json).

PR #213, opened 2021-07-16
fix: correctly validate token

drazisil-codecov:


thomasrockhu (approved):
Dat formatting though

RA80533:
> > Dat formatting though
> 
> I don't even know what formatter that is anymore. Every time I blink I swear VSCode changes it on me.

Related #210

RA80533:
Per https://github.com/codecov/uploader/pull/212#issuecomment-881623245:

> The token is exposed as part of the destination for the HTTP/HTTPS request being sent. Since the destinations must be parsed in order to route them along to the destination itself, each node along the route a given request takes will see the token. This happens even if the request is sent using HTTPS.
> 
> /cc @drazisil @drazisil-codecov @vlad-ko

RA80533:
If a source-level modification is required for operation to occur, should the change be made at the backend in such a way to not require this vulnerability to exist?

RA80533:
> The token is only able to upload coverage reports, nothing else, so it is not a sensitive as an API token.

Ah, I believe I understand. Just to reiterate, the token is associated with a specific coverage report and is not an authoritative credential of the user uploading the report? In other words, if I were to send a request with the token as it is logged in the CI, there would be additional authentication measures in place to prevent me from impersonating the uploader?

PR #221, opened 2021-07-19
refactor: add getToken()

drazisil-codecov:
Add getToken() helper to ensure correct token is set
Configure typescript linter and add missing types


thomasrockhu (approved):

tsconfig.json
```diff
@@ -0,0 +1,64 @@
+{
```
thomasrockhu, line 1: this... could be cleaner

drazisil-codecov: It could. That was the way the formatter made it. I'll kick it.

Issue #155, opened 2021-06-21
`isBlacklisted()` does not work for glob patterns


RA80533:

drazisil-codecov:
Line https://github.com/codecov/uploader/blob/master/src/helpers/files.js#L179 will likely need glob to expand the masks.

RA80533:
It appears that the glob library this project uses calls minimatch under the hood for this functionality. minimatch could be added as a direct dependency for its [match](https://www.npmjs.com/package/minimatch#minimatchmatchlist-pattern-options) functionality.

drazisil-codecov:
> It appears that the glob library this project uses calls minimatch under the hood for this functionality. minimatch could be added as a direct dependency for its [match](https://www.npmjs.com/package/minimatch#minimatchmatchlist-pattern-options) functionality.
> 
> minimatch does appear to be quite old; however:
> 
> * It appears to be quite mature
> * It is written by one of the founders of npm (the former [BDFL](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life), in fact)

So is Glob :D. I'm not adverse to the change, at the time I was the only dev and I understand glob, not so much how to use minimatch. I'd love to look at a PR, since it's not broken right now it's going to be lower priority If I need to do it

mitchell-codecov:
Fixed in #276

Issue #152, opened 2021-06-21
I/O is synchronous


RA80533:

drazisil-codecov:
Yesm I went that route because the process has to wait to complete anyway. I'm open to improvements though! :)

drazisil-codecov:
This was part of your gob improvement PR, right, @RA80533 ?

RA80533:
Yep!

drazisil:
@mitchell-codecov this is no longer the case, I believe

Issue #98, opened 2021-05-24
tests should not rely on environmental state


drazisil-codecov:
**Describe the bug**

Unit tests need to be isolated. The branch detection tests will fail if git is not installed. 

**To Reproduce**
1. Run npm test on a machine that does not have git installed.

**Expected behavior**

Test should use passed envs and not be affected by outside state.

**Screenshots**

**Additional context**

Issue #36, opened 2020-10-01
Scan paths and suggest path fixes


drazisil-codecov:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
If paths in the report and the repo do not match, would love a suggested path fix

**Describe alternatives you've considered**
Manually creating path fixes

**Additional context**
N/A

thomasrockhu-codecov:
lolol, out of scope now @drazisil-codecov?

PR #224, opened 2021-07-20
Improve glob and file operations

drazisil-codecov:
Currently the ignore feature does not work, and the blacklist is static.
After adding test guards, will optimize directory searching and add ignore feature.

Related:
- #155 
- #98 
- #152 
- #36


drazisil:
Ping @drazisil-codecov

PR #225, opened 2021-07-20
Fix Jenkins slug and linting

thomasrockhu:
- #218 
- #220
- #221


RA80533:
I'm starting to see the release process a bit more clearer. Merge feature branches into `master` then merge `master` into `release` to trigger an asset build and release on GitHub?

src/index.js
```diff
@@ -243,7 +239,7 @@ async function main(args) {
       { level: 'debug', args },
     )
     const result = await webHelpers.uploadToCodecovPUT(uploadURL, gzippedFile)
-    log(result)
+    log(JSON.stringify(result))
```
drazisil-codecov, line 242: I'm really starting to wish I went with a logger like winston then rolling my own logger. Needs the balance between sanity and file size. 🤔

test/providers/provider_gitlabci.test.js
```diff
@@ -47,6 +47,14 @@ describe('GitLabCI Params', () => {
       service: 'gitlab',
       slug: '',
     }
+    const spawnSync = td.replace(childProcess, 'spawnSync')
```
drazisil-codecov, line 50: I think we can safely replace this with `execFileSync` in the future, since we are no longer calling both `git` and `hg`

drazisil-codecov (approved):

Issue #222, opened 2021-07-20
Error while uploading to Codecov (Not Found)


jlimas:
We started getting the following error message on our Bitbucket Pipelines.

Detected Bitbucket as the CI provider.
Pinging Codecov: https://codecov.io/upload/v4?package=uploader-0.1.0&token=*******....................
Error!: Error: Error uploading to https://codecov.io: Error: Error uploading to Codecov: Error: Not Found

Didn't see any incident on your website regarding this.

TriplEight:
I'm getting the same: 
<details>

```
 codecov -t ******* -f lcov-lines.info -v

     _____          _
    / ____|        | |
   | |     ___   __| | ___  ___ _____   __
   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
   | |___| (_) | (_| |  __/ (_| (_) \ V /
    \_____\___/ \__,_|\___|\___\___/ \_/

  Codecov report uploader 0.1.0
=> Project root located at: /builds
Start of network processing...
Searching for files in /builds

....

Search complete for files in /builds
End of network processing
Processing lcov-lines.info...
Detected Local as the CI provider.
Pinging Codecov: https://codecov.io/upload/v4?package=uploader-0.1.0&token=*******&branch=master&build=&build_url=&commit=e5b400bee45556d948cf0e818f9d8ffc016fe09b&job=&pr=&service=&slug=paritytech/ink&name=&tag=&flags=&parent=
https://codecov.io/upload/v4?package=uploader-0.1.0&branch=master&build=&build_url=&commit=e5b400bee45556d948cf0e818f9d8ffc016fe09b&job=&pr=&service=&slug=paritytech/ink&name=&tag=&flags=&parent=
        Content-Type: 'text/plain'
        Content-Encoding: 'gzip'
        X-Reduced-Redundancy: 'false'
Error!: Error: Error uploading to https://codecov.io: Error: Error uploading to Codecov: Error: Bad Request
```

thomasrockhu:
@jlimas I made an update to the uploader, would you mind trying to see if you are still experiencing the issue?

jlimas:
Hello @thomasrockhu thanks for the follow up, I'm doing a test right now downloading the tool with this command and the env has the CODECOV_TOKEN setup correctly.

`curl -Os https://uploader.codecov.io/latest/linux/codecov`
`./codecov -t ${CODECOV_TOKEN} -f reports/workflows-interpreter.xml`

[...]

Issue #219, opened 2021-07-19
The Codecov uploader doesn't work in Docker containers that runs in TeamCity with corporate proxy


anmyachev:
**Describe the bug**
```sh
Pinging Codecov: https://codecov.io/upload/v4?package=uploader-0.1.0&token=*******&branch=undefined&build=&build_url=&commit=75607b5fa42ddbaf29f3cc80593de9a1c71a9b6b&job=&pr=&service=teamcity&slug=&name=&tag=&flags=&parent=
Error!: Error: Error uploading to https://codecov.io: Error: Error uploading to Codecov: Error: connect ETIMEDOUT 34.120.215.12:443
```

**To Reproduce**
1. Download codecov from `curl -Os https://uploader.codecov.io/latest/linux/codecov`
2. Run docker container in TeamCity using codecov env script
  ```sh
  # request codecov to detect CI environment to pass through to docker
ci_env=`bash <(curl -s https://codecov.io/env)`
docker run $ci_env -e CI=true ...
   |----- inside docker container
   | # exec tests
   | # Your Codecov Uploader approach here
   | # See more at https://docs.codecov.com/docs/codecov-uploader
   | ./codecov
  ```

**Expected behavior**
Bash uploader works well in the case.

Judging by parameters of the query string and by `https://github.com/codecov/uploader/blob/master/src/ci_providers/provider_teamcity.js` file, the new uploader does not work with TeamCity variables defined by env script:
```
elif [ "$TEAMCITY_VERSION" != "" ];
then
  add "TEAMCITY_VERSION"
  add "TEAMCITY_BUILD_BRANCH"
  add "TEAMCITY_BUILD_ID"
  add "TEAMCITY_BUILD_URL"
  add "TEAMCITY_BUILD_COMMIT"
  add "TEAMCITY_BUILD_COMMIT"
  add "BUILD_VCS_NUMBER"
  add "TEAMCITY_BUILD_REPOSITORY"
```

Version of env script - `1.0.3` (Also note, that line `add "TEAMCITY_BUILD_COMMIT"` is repeated twice - maybe typo)

thomasrockhu:
@anmyachev I released a fix that should be deployed in the next hour or so, would you be able to see if it works for you?

anmyachev:
@thomasrockhu Thank you! I tried a newer version. It doesn't work yet, but the error looks a little different.
From the last time, `slug` parameter began to be correctly determined.
[...]

Issue #217, opened 2021-07-17
The uploader receives bad requests when setting CI env in GitHub actions


dustin71728:
**Describe the bug**
When I run the codecov uploader in the docker container, everything is fine until I set up environment variables using the first technique mentioned in the following article:
https://docs.codecov.com/docs/testing-with-docker

Then the uploader complains about bad requests:
```
Detected GitHub Actions as the CI provider.
Pinging Codecov: https://codecov.io/upload/v4?package=uploader-0.1.0&token=*******&branch=test/migrate-to-shellspec&build=1039293011&build_url=undefined%2Fdustin71728%2Fyarn-2-completion%2Factions%2Fruns%2F1039293011&commit=24eb740b7574622e2af937f65c74e97292ab9448&job=default&pr=&service=github-actions&slug=dustin71728/yarn-2-completion&name=&tag=&flags=&parent=
https://codecov.io/upload/v4?package=uploader-0.1.0&branch=test/migrate-to-shellspec&build=1039293011&build_url=undefined%2Fdustin71728%2Fyarn-2-completion%2Factions%2Fruns%2F1039293011&commit=24eb740b7574622e2af937f65c74e97292ab9448&job=default&pr=&service=github-actions&slug=dustin71728/yarn-2-completion&name=&tag=&flags=&parent=
        Content-Type: 'text/plain'
        Content-Encoding: 'gzip'
        X-Reduced-Redundancy: 'false'
Error!: Error: Error uploading to https://codecov.io: Error: Error uploading to Codecov: Error: Bad Request
```
I wonder if the root cause is that the build_url query parameter is not a well-formatted URL. It is a  literal undefined string in the host part of the URL.

So I try to set up CI_BUILD_URL with the correct value, but it still doesn't work. 
The failed log( In `Run yarn test` step):
https://github.com/dustin71728/yarn-2-completion/runs/3091016225?check_suite_focus=true

The log printed out all the environment variables returned from the command:
"ci_env=`bash <(curl -s https://codecov.io/env)`"
https://github.com/dustin71728/yarn-2-completion/runs/3090999690?check_suite_focus=true

The successful log without CI environment variables involved:
https://github.com/dustin71728/yarn-2-completion/runs/3081701233?check_suite_focus=true

**References**
Dockerfile: https://github.com/dustin71728/yarn-2-completion/blob/test/migrate-to-shellspec/docker/bash-5-0.Dockerfile
[...]

PR #227, opened 2021-07-21
feat: Allow for token from yaml

thomasrockhu:
fixes https://github.com/codecov/codecov-action/issues/430
related https://github.com/codecov/uploader/issues/222
related https://github.com/codecov/uploader/issues/219
related https://github.com/codecov/uploader/issues/217

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
     "glob": "7.1.7",
+    "js-yaml": "4.1.0",
     "line-reader": "0.4.0",
     "superagent": "6.1.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
     "glob": "7.1.7",
+    "js-yaml": "4.1.0",
```
drazisil-codecov, line 31: If we are parsing the codecov.yml, we should consider the `ignore` section when we implement the correct igmore functionality. If the customer doesn't want certain paths reported on, those paths probably don't contain coverage files.

drazisil-codecov (approved):

Issue #229, opened 2021-07-21
Local provider fails if `git` is not available


dbhobbs:
I'm in the process of building a container to use as a Google Cloud Builder for uploading coverage reports to codecov. This container does not have `git` available as all arguments are passed as environment variables. During execution the uploader fails with: `TypeError: Cannot read property 'toString' of null`

Full output of the run:
```

     _____          _
    / ____|        | |
   | |     ___   __| | ___  ___ _____   __
   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
   | |___| (_) | (_| |  __/ (_| (_) \ V /
    \_____\___/ \__,_|\___|\___\___/ \_/

  Codecov report uploader 0.1.0
=> Project root located at: /workspace
=> Found 1 possible coverage files:
coverage.out
Detected Local as the CI provider.
Error!: Error: There was an error getting the branch name from git: TypeError: Cannot read property 'toString' of null
```

**Expected behavior**
I would not expect `git` to be a required application. Particularly in the case where the value is provided in the argument list and the `git` provided value isn't used anyway.

thomasrockhu:
@dbhobbs I pushed a fix for this that should be live in the next hour, would you be able to see if it fixes the problem?

dbhobbs:
@thomasrockhu I rebuilt the container, but I didn't notice any difference and I'm still seeing the `toString` error. That could be due to some caching, but it's hard to know what version of the uploader we're running as they're all `version: 0.1.0`. Can we maybe add the build number as well as the version number to the uploader output?

PR #231, opened 2021-07-21
fix: Try to use args before git

thomasrockhu:
Fixes https://github.com/codecov/uploader/issues/229


dbhobbs (commented):
Should the if statements also guard against an empty string?

thomasrockhu:
@dbhobbs, I think `''` should fall out of the `if` statement, is that what you mean?

dbhobbs:
Oh right... javascript treats empty strings as truthy. :blush:

thomasrockhu:
@dbhobbs correct

drazisil-codecov (approved):

PR #235, opened 2021-07-22
Better logs

drazisil-codecov:
Create distinct log methods with levels and timestamps
Add some verbose logs and set the stage to easily add more

### DIFF
--- a/bin/codecov
+++ b/bin/codecov
@@ -115,22 +115,16 @@ var argv = require('yargs') // eslint-disable-line
 
 const start = Date.now()
 
-app.log(`Start of uploader: ${start}...`, { level: 'debug', argv })
+app.verbose(`Start of uploader: ${start}...`, argv.verbose)
 app
   .main(argv)
   .then(() => {
     const end = Date.now()
-    app.log(`End of uploader: ${end - start} milliseconds`, {
-      level: 'debug',
-      argv,
-    })
+    app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
   })
   .catch(error => {
     app.log(`Error!: ${error}`, { level: 'error', argv })
     const end = Date.now()
-    app.log(`End of uploader: ${end - start} milliseconds`, {
-      level: 'debug',
-      argv,
-    })
+    app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
     process.exit(argv.nonZero ? -1 : 0)
   })
--- a/bin/codecov
+++ b/bin/codecov
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-const { error, main, verbose} = require('../src')
+const app = require('../src')
 
 var argv = require('yargs') // eslint-disable-line
   .usage('Usage: $0 <command> [options]')
@@ -106,7 +106,7 @@ var argv = require('yargs') // eslint-disable-line
       alias: 'Q',
       type: 'string',
       description: `Used internally by Codecov, this argument helps track wrappers
-        of the uploader (e.g. GitHub Action, CircleCI Orb)`
+        of the uploader (e.g. GitHub Action, CircleCI Orb)`,
     },
   })
   .version()
@@ -115,15 +115,22 @@ var argv = require('yargs') // eslint-disable-line
 
 const start = Date.now()
 
-verbose(`Start of uploader: ${start}...`, argv.verbose)
-main(argv)
+app.log(`Start of uploader: ${start}...`, { level: 'debug', argv })
+app
+  .main(argv)
   .then(() => {
     const end = Date.now()
-    verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
+    app.log(`End of uploader: ${end - start} milliseconds`, {
+      level: 'debug',
+      argv,
+    })
   })
-  .catch(err => {
-    error(`Error!: ${err}`)
+  .catch(error => {
+    app.log(`Error!: ${error}`, { level: 'error', argv })
     const end = Date.now()
-    verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
+    app.log(`End of uploader: ${end - start} milliseconds`, {
+      level: 'debug',
+      argv,
+    })
     process.exit(argv.nonZero ? -1 : 0)
   })
--- a/bin/codecov
+++ b/bin/codecov
@@ -122,8 +122,8 @@ app
     const end = Date.now()
     app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
   })
-  .catch(error => {
-    app.error(`Error!: ${error}`)
+  .catch(err => {
+    app.error(`Error!: ${err}`)
     const end = Date.now()
     app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
     process.exit(argv.nonZero ? -1 : 0)
--- a/bin/codecov
+++ b/bin/codecov
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-const app = require('../src')
+const { error, main, verbose} = require('../src')
 
 var argv = require('yargs') // eslint-disable-line
   .usage('Usage: $0 <command> [options]')
@@ -115,16 +115,15 @@ var argv = require('yargs') // eslint-disable-line
 
 const start = Date.now()
 
-app.verbose(`Start of uploader: ${start}...`, argv.verbose)
-app
-  .main(argv)
+verbose(`Start of uploader: ${start}...`, argv.verbose)
+main(argv)
   .then(() => {
     const end = Date.now()
-    app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
+    verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
   })
   .catch(err => {
-    app.error(`Error!: ${err}`)
+    error(`Error!: ${err}`)
     const end = Date.now()
-    app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
+    verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
     process.exit(argv.nonZero ? -1 : 0)
   })
--- a/bin/codecov
+++ b/bin/codecov
@@ -123,7 +123,7 @@ app
     app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
   })
   .catch(error => {
-    app.log(`Error!: ${error}`, { level: 'error', argv })
+    app.error(`Error!: ${error}`)
     const end = Date.now()
     app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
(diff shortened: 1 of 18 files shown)


src/index.js
```diff
@@ -146,13 +146,13 @@ async function main(args) {
...
-      log(`Processing ${coverageFile}...`),
+      info(`Processing ${coverageFile}...`),
         (fileContents = await fileHelpers.readCoverageFile(
           args.dir || projectRoot,
           coverageFile,
         ))
-    } catch (error) {
-      log(`Could not read coverage file (${coverageFile}): ${error}`)
+    } catch (err) {
+      info(`Could not read coverage file (${coverageFile}): ${err}`)
```
drazisil-codecov, line 155: @thomasrockhu I think this should be an error, but I didn't want to redo the tests. Also, if this is an error, why do we continue? Can we reword this to _potential_ coverage file? If we tell a customer we can't read a coverage file, that should be an error, in my opinion.

src/helpers/logger.js
```diff
@@ -1,7 +1,21 @@
...
+ * * Info
+ * * Verbose
+ * For the purposes of minimal refactor, I'm aliasing debug() to verbose()
+ * I'm also mapping the log() options to the new calls and
+ * deprecating log()
+ */
+
+function _getTimestamp() {
+  return new Date().toISOString()
+}
```
RA80533, line 15: 🙏

src/ci_providers/provider_azurepipelines.js
```diff
@@ -1,5 +1,5 @@
 const childProcess = require('child_process')
-const { log } = require('../helpers/logger')
+const { error, info, verbose } = require('../helpers/logger')
```
thomasrockhu, line 2: does it need `error` or `verbose`?

src/ci_providers/provider_githubactions.js
```diff
@@ -1,5 +1,5 @@
 const childProcess = require('child_process')
-const { log } = require('../helpers/logger')
+const { error, info, verbose } = require('../helpers/logger')
```
thomasrockhu, line 2: same, only uses `info`

src/ci_providers/provider_githubactions.js
```diff
@@ -70,10 +70,10 @@ function _getSHA(inputs) {
       .trimRight()
     if (mergeCommitRegex.exec(mergeCommitMessage)) {
       const mergeCommit = mergeCommitMessage.split(' ')[1]
-      log(`    Fixing merge commit SHA ${commit} -> ${mergeCommit}`)
+      info(`    Fixing merge commit SHA ${commit} -> ${mergeCommit}`)
       commit = mergeCommit
     } else if (mergeCommitMessage === '') {
-      log(
+      info(
```
thomasrockhu, line 76: could be an error..., we can leave it for now

src/helpers/token.js
```diff
@@ -2,33 +2,29 @@ const fs = require('fs')
...
+const { error, info, verbose } = require('./logger')
 const validateHelpers = require('./validate')
 
 function getToken(inputs, projectRoot) {
   const { args, envs } = inputs
   const options = [
     [args.token, 'arguments'],
     [envs.CODECOV_TOKEN, 'environment variables'],
-    [getTokenFromYaml(projectRoot), 'Codecov yaml config'],
+    [getTokenFromYaml(projectRoot, args), 'Codecov yaml config'],
```
thomasrockhu, line 13: this doesn't use `args`

src/helpers/token.js
```diff
@@ -2,33 +2,29 @@ const fs = require('fs')
...
   return ''
 }
 
-function getTokenFromYaml(projectRoot) {
-  const dirNames = [
-    '',
-    '.github',
-    'dev',
-  ]
+function getTokenFromYaml(projectRoot, args) {
```
thomasrockhu, line 26: `args` is not used

src/index.js
thomasrockhu: yeah I agree that it should be `potential`

src/index.js
```diff
@@ -210,13 +210,14 @@ async function main(args) {
     return dryRun(uploadHost, token, query, uploadFile, args.source)
   }
 
-  log(
+  info(
     `Pinging Codecov: ${uploadHost}/upload/v4?package=${webHelpers.getPackage(
       args.source,
     )}&token=*******&${query}`,
   )
+  verbose(`Passed token was ${token.length} characters long`, args.verbose)
```
thomasrockhu, line 218: 👏

thomasrockhu: We might want to put this in `step 3`, though to keep `token` things together

thomasrockhu (approved):
Mostly nits

src/helpers/token.js
drazisil-codecov: It does, on line 55

RA80533:
Is it in the cards to bring the CI over to GitHub Actions?

RA80533:
Does the CI build the binary it's trying to run or is it the latest public version?

RA80533:
Just out of curiosity, what ended up being the issue?

RA80533:
```console
$ node bin/codecov
```
```console
/[…]/uploader/bin/codecov:126
    app.log(`Error!: ${error}`, { level: 'error', argv })
        ^

TypeError: app.log is not a function
    at /[…]/uploader/bin/codecov:126:9
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

RA80533:
What editor do you use? I don't believe my set-up is anything extraordinary but VS Code shows me a handful of errors throughout the source code.

RA80533:
> > What editor do you use? I don't believe my set-up is anything extraordinary but VS Code shows me a handful of errors throughout the source code.
> 
> VSCode. It's being silly and not highlighting and I'm too flustered to troubleshoot it. Thank you for the catch. :sigh:

Yeah, sometimes the extensions fail :shrug: The ESLint extension used to be so unreliable for me in that regard that I just switched over entirely to running the CLI. Sometimes I would get the lines, sometimes the issues were reported twice, sometimes they weren't reported at all.

I just remembered that I'm on VS Code Insiders so whatever editing experience I'm having is probably going to be different from the stable releases if that's what you're on.

src/index.js
```diff
@@ -199,6 +199,7 @@ async function main(args) {
 
   if (serviceParams === undefined) {
     logAndThrow('Unable to detect service, please specify manually.')
+    process.exit(args.nonZero ? -1 : 0)
```
RA80533, line 202: If the previous line actually throws an error, this line won’t actually execute. The functionality should be in ../bin/codecov, if I’m not mistaken, however.

RA80533:
Gotcha. You can add `"checkJs": true` to `"compilerOptions"` in tsconfig.json to make it behave.

src/index.js
drazisil-codecov: No, I agree. `npm run type-check` was mad that it wasn't returning since it doesn't know that function throws. I decided I preferred extra lines that never get hit over fighting.

RA80533: That makes sense.

TypeScript is normally able to figure that out but it may have difficulty right now because everything is in JS. Written in TS, it would see a function that always throws an error and implicitly type it as returning `never`: 

```ts
function defcon(level: number): never {
  throw new Error(`DEFCON ${level}`);
}
```

Adding the corresponding `@returns` annotation to `logAndThrow`’s JSDoc comment might provide TypeScript with the information it needs to satisfy it requiring return statements:
```js
/** @returns {never} */
```

drazisil-codecov: What is the difference between void and never, in this case? Iy's normally a healthy void.

bin/codecov
```diff
@@ -115,22 +115,16 @@ var argv = require('yargs') // eslint-disable-line
...
     const end = Date.now()
-    app.log(`End of uploader: ${end - start} milliseconds`, {
-      level: 'debug',
-      argv,
-    })
+    app.verbose(`End of uploader: ${end - start} milliseconds`, argv.verbose)
   })
   .catch(error => {
-    app.log(`Error!: ${error}`, { level: 'error', argv })
+    app.error(`Error!: ${error}`)
```
RA80533, line 126: Is `error` exported in ../src/index.js?

drazisil-codecov: @*#. No.

src/index.js
RA80533: Good question.

- A return type of `never` signifies that the function will never run to completion. Functions that always throw an error or ones that terminate the program (e.g., `process.exit`) are typed as returning `never`.
- A return type of `void` means the function does not return a (defined) value. In practice, this just means the function evaluates to `undefined`, usually because the function has no return statements or the return statements it has are empty.

RA80533:
> > "checkJs": true
> 
> Ah, missed that one, thanks :)

Yup. A recent change (TypeScript 4.3 or 4.4, I forget) made `checkJs` imply `allowJs` so it’s now the one to use.

bin/codecov
drazisil-codecov: > > > "checkJs": true
> > 
> > 
> > Ah, missed that one, thanks :)
> 
> Yup. A recent change (TypeScript 4.3 or 4.4, I forget) made `checkJs` imply `allowJs` so it’s now the one to use.

I read that, and mentally reversed them, which is why I only had the one

RA80533:
> https://github.com/vercel/pkg/blob/5.1.0/prelude/bootstrap.js#L1491-L1502
> 
> https://nodejs.org/docs/latest-v16.x/api/vm.html#vm_class_vm_script
> 
> https://nodejs.org/docs/latest-v16.x/api/vm.html#vm_script_runinthiscontext_options

The error is sourced there because that's where the footprints end so to speak. What's happening is the JavaScript is being packaged as V8 bytecode which is then executed through the `vm` module. The V8 virtual machine encounters the `TypeError: undefined is not a function` error while interpreting the bytecode.

```console
$ NODE_DEBUG=vm out/codecov-alpine […] >> output_alpine.txt
```

Is output_alpine.txt uploaded anywhere as an artifact from the run?

src/helpers/token.js
```diff
@@ -5,6 +5,12 @@ const yaml = require('js-yaml')
 const { error, info, verbose } = require('./logger')
 const validateHelpers = require('./validate')
 
+/**
+ *
+ * @param {object} inputs
+ * @param {strinh} projectRoot
```
RA80533, line 11:
```suggestion
 * @param {string} projectRoot
```

😸

bin/codecov
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-const app = require('../src')
+const { error, main, verbose} = require('../src')
```
RA80533, line 3: 🙏 This syntax allows for the most optimization for bundling.

RA80533:
It might be helpful to check to see if the original source code runs.

RA80533:
> I still don't know [11b7bcc](https://github.com/codecov/uploader/commit/11b7bcc133187e51034080e2e45e6a871e97ace3) fixed it, but I'm merging this before I lose a working status

I'm right with you there.

PR #244, opened 2021-07-23
test: alpine, a decent into madness

drazisil-codecov:
A investigation into why the parent of https://github.com/codecov/uploader/commit/11b7bcc133187e51034080e2e45e6a871e97ace3 causes Alpine to fail to execute the binary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ jobs:
       - run:
           name: Run Alpine binary -f (dry run)
           command: |
-            NODE_DEBUG=vm out/codecov-alpine -f /root/project/coverage-alpine/cobertura-coverage.xml -F alpine -d -Z -e CIRCLE_BRANCH >> output_alpine.txt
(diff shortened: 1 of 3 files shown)


RA80533:
See #243

RA80533:
A few thoughts:

- Bytecode compilation through ncc is not deterministic.
- Some non-working binary may have been cached.

It might be ideal to perform simple checksums to determine if a file matches up between CI runs.

RA80533:
> > #243
> 
> What about it?

It reverts the commit that made the other PR work; however, it does not cause any of the previous issues.

RA80533:
> I believe you reverted the wrong commit.

Good catch.

RA80533:
> Completely unrelated, I discovered last night that building a Windows installer in Node is _hard_

I can imagine. I had the pleasure of creating a Java-based installer for Windows many moons ago and even the heavy abstractions were an annoyance. The Windows API was not in a great shape then either—I believe it was during the awkward transition between 7 and 8.

PR #249, opened 2021-07-23
Optimize glob functionality

RA80533:
This PR makes globbing asynchronous.

- Test runs in the CI are ~1,450% faster

### Before
https://app.circleci.com/pipelines/github/codecov/uploader/887/workflows/9076a1ff-104c-49cb-b944-769380335b6e/jobs/6518/parallel-runs/0/steps/0-105
> 1m 49s

### After
https://app.circleci.com/pipelines/github/codecov/uploader/906/workflows/a490529d-ad28-4c3b-9c64-6eb5a707f96f/jobs/6718/parallel-runs/0/steps/0-105
> 7s

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
+    "fast-glob": "^3.2.7",
     "glob": "7.1.7",
     "js-yaml": "4.1.0",
     "line-reader": "0.4.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "license": "ISC",
       "dependencies": {
+        "fast-glob": "^3.2.7",
         "glob": "7.1.7",
         "js-yaml": "4.1.0",
         "line-reader": "0.4.0",
@@ -1254,7 +1255,6 @@
       "version": "2.1.4",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.4.tgz",
       "integrity": "sha512-33g3pMJk3bg5nXbL/+CY6I2eJDzZAni49PfJnL5fghPTggPvBd/pFNSgJsdAgWptuFu7qq/ERvOYFlhvsLTCKA==",
-      "dev": true,
       "dependencies": {
         "@nodelib/fs.stat": "2.0.4",
         "run-parallel": "^1.1.9"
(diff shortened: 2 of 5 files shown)


RA80533:
No change was made to provider_circleci.js nor any of its immediate dependencies. The drop in coverage doesn't seem accurate. Istanbul reports the file to have full coverage.

thomasrockhu:
@RA80533 would love to know how you ended up figuring this out and why there's such a dramatic difference in time

package.json
```diff
@@ -27,6 +27,7 @@
   },
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
+    "fast-glob": "^3.2.7",
```
drazisil-codecov, line 30: Nice!
Wasn't it you though that said BDL's libs were perfect? :trollface:

drazisil-codecov (approved):
How much faster is it on a live run? I'm using a mocked filesystem for tests that has 5 files...I'd expect it to be fast.

package.json
RA80533: Haha, yeah. I figured if this was going to be a speed improvement it would be a *speed improvement*. I believe it uses good ol' minimatch under the hood, however.

RA80533:
> @RA80533 would love to know how you ended up figuring this out and why there's such a dramatic difference in time

Synchronous IO blocks execution until data is received/sent. Something like directory traversal does a lot of trivial requests to the OS to access IO (the file system, in this case) which would cause the program to run in somewhat of a stutter. When this is asyncronous, many of the calls can be sent all at once and other ones can weave in and out of the calls being received, etc.

Most of the speed improvement came from the asynchrony. fast-glob was used in place of glob because it is more optimized for handling a lot of glob patterns rather than a few (for which glob tends to be faster).

package.json
drazisil-codecov: I think I saw `micromatch`, but I've already forgot. Code bits like https://github.com/codecov/uploader/issues/251 are making me question why I woke up at 4am today

drazisil-codecov (commented):
Quick thought here, @RA80533 
Can you double-check me that we aren't walking the tree twice, once for file listing and once for coverage file search?

My fuzzy 🧠 is making it hard for me to tell (and I'm focused in other things right now)

RA80533:
I think I know what you're taking about.

https://github.com/codecov/uploader/blob/51849d0da8af47d65cb66964384887100728f72c/src/helpers/files.js#L232

This function is called in addition to the coverage file search. It has room for optimization on its own but there could be further optimization by using it to obtain a directory listing *then* running the glob functionality to obtain the list of coverage files used elsewhere.

PR #256, opened 2021-07-26
fix: Token should be codecov.token and update symlinks

thomasrockhu:
- Tokens should be pulled from `codecov.token` not `codecov_token`
- Symlinks that point to non-existing files should not error on network pull
- Regenerate `npm-shrinkwrap`

fixes https://github.com/codecov/uploader/issues/238
https://github.com/codecov/codecov-action/issues/446
https://github.com/codecov/codecov-action/issues/445
https://github.com/codecov/codecov-action/issues/442
https://github.com/codecov/codecov-action/issues/435


thomasrockhu:
@RA80533 

1. `codecov_token` is the wrong config field, it should have been `codecov.token`
2. symlinks to files that don't exist breaks on `fs.stat*`. Using `fs.lstat` instead
3. just re-running `npm shrinkwrap`

RA80533:
@thomasrockhu It looks like you're using an older version of npm. Could you try to run `npm --version`? npm v7 generates lock file version 2 which is what the project is using.

drazisil-codecov (approved):
Outside of FOSSA, looks good.

thomasrockhu:
@RA80533 you would be accurate, I upgraded and pushed a new version

RA80533:
Although it's not part of the changes in this PR, line 249 looks to be the source of a future bug:

https://github.com/codecov/uploader/blob/a112caa9c9ee0e08964fea781177c6c3e80a3cc8/src/helpers/files.js#L249

```diff
- `${path.join(dirPath.replace(projectRoot, '.'), file)}\n`, 
+ `${path.join(dirPath, file)}\n`,
```

thomasrockhu:
@drazisil @RA80533 any idea why `husky` is not cooperating?

thomasrockhu:
```
--> npm -v
7.20.1
```

RA80533:
I suspect it's npm-shrinkwrap.json. I'm not sure why there are differences in the likes of a few properties being removed. If it needs to be regenerated during installation, the installation isn't deterministic which is sort of what we're seeing here.

RA80533:
- [path.relative](https://nodejs.org/api/path.html#path_path_relative_from_to)
- [path.normalize](https://nodejs.org/api/path.html#path_path_normalize_path) (if necessary)

RA80533:
@thomasrockhu Could you transplant the old npm-shrinkwrap.json to the PR and see if the CI accepts it, etc.? Would that version of npm-shrinkwrap.json be modified via `npm install` on your system?

thomasrockhu:
@RA80533 @drazisil-codecov so strange, I pulled the version on `master` and it looks roughly the same (just a tiny lint change) when running `npm install` locally.

RA80533:
> Do you read the Node docs cover to cover? I'm not making fun, I'm very impressed on how well you know them.

Haha, you're fine. In this instance, I did not know the specific functions I was referring to off the top of my head but I knew they existed. I suppose it's the difference between not knowing what you don't know versus knowing what you don't know.

PR #273, opened 2021-08-04
refactor: use async file read, use mock-fs to test

drazisil:
* use promise-based read file
* use mock-fs for fake filesystem, instead of mocking node module


.circleci/config.yml
```diff
@@ -159,13 +159,20 @@ jobs:
       - attach_workspace:
           # Must be absolute path or relative path from working_directory
           at: .
+      # https://superuser.com/a/1620000/54442
+      - run:
+          name: Install and set Node 16.x
+          command: |
+            $ErrorActionPreference="Stop"
```
mitchell-codecov, line 166: Are the lines below treated as if they were semicolon-separated?

.circleci/config.yml
```diff
@@ -159,13 +159,20 @@ jobs:
...
+            nvm arch
+            nvm install 16 all
+            nvm use 16
```
mitchell-codecov, line 169: Where is nvm obtained? Does CircleCI have it by default?

src/helpers/files.ts
```diff
@@ -293,18 +294,15 @@ export function readAllLines(filePath: string): string[] {
...
-  } catch (error) {
-    logAndThrow(`There was an error reading the coverage file: ${error}`)
-  }
-  return ''
+): Promise<string> {
+  return readFile(getFilePath(projectRoot, filePath), {
+    encoding: 'utf-8',
+  }).catch(err => {
+    logAndThrow(`There was an error reading the coverage file: ${err}`)
+  })
```
mitchell-codecov, line 305:
```suggestion
  })
```

mitchell-codecov (approved):

.circleci/config.yml
drazisil-codecov: Ish. They are separate commands on seperate lines :)

.circleci/config.yml
drazisil-codecov: Yes, the image has https://github.com/coreybutler/nvm-windows installed

src/helpers/files.ts
drazisil-codecov: Why?

.circleci/config.yml
mitchell-codecov: Cool. I wasn't sure if CircleCI had any magic juice it used to run those commands.

drazisil-codecov: If you consider PowerShell 5 'magic' 😆

.circleci/config.yml
mitchell-codecov: Do you know if that dependency is something we configured?

src/helpers/files.ts
mitchell-codecov: It's the caller's responsibility to handle errors / catch and log them. This is actually what appears to be happening right now so the change would not introduce incompatibility of anything that calls it:

https://github.com/codecov/uploader/blob/7567c5e5c6586ad9ddf151b1807c3bba629daca3/src/index.ts#L182

mitchell-codecov: Additionally, the catch-and-throw pattern, when used this way, creates nested errors that are wrapped several times (which is why we're seeing "Error: Received an Error: Received an Error: Received an unexpected response" and the like).

drazisil-codecov: Ah. Sorry I didn't see your response before merge

Issue #259, opened 2021-07-27
Tox and __pycache__ files should be skipped


avylove:
**Describe the bug**
When uploading reports for a Python project, the `.tox` and `__pycache__` directories are include in the file network. These are both generated by the build and won't contain any source files. This can create issues with what files are listed on the website and is preventing us from using the uploader.

**To Reproduce**
1. Test and run coverage on a python project with tox
2. Use uploader 
3. Look at file list

**Expected behavior**
`.tox` and `__pycache__` directories should be filtered out when files network is generated

**Screenshots**
[example](https://codecov.io/api/gh/jquast/blessed/download/build?path=v4/raw/2021-07-26/E86BE245FC61F0200038303B95EC2F5B/a679854dea6691a82f3fbd12736150df90280ca2/a82bc55c-61b8-4e8e-a2c9-e98b1b893862.txt)

**Additional context**
`.tox` contains the virtual environments created by the tox utility to run tests
`__pycache__` contains Python bytecode files

PR #280, opened 2021-08-04
fix: Remove __pycache__ and .tox folders from search

thomasrockhu-codecov:
I'll be honest, not totally sure how we are doing `manualBlacklist` and `globBlacklist` anymore. 

fixes https://github.com/codecov/uploader/issues/259


drazisil-codecov (approved):
I think it's only needed in one of them. They are combined now and passed to glob.

PR #291, opened 2021-08-11
Add new Apex file format to upload ignore list

michael-codecov:
Salesforce has added new json files `test-result-<testrunid>-codecoverage.json` that contain class/method coverage info on a per test basis. We only need the existing `test-result-codecoverage.json` file to calculate per line coverage, so we can safely ignore the other ones.


drazisil-codecov (changes requested):
Disread this, correct review below

drazisil-codecov (approved):
This looks like a good solution on the uploader side

PR #296, opened 2021-08-11
Don't upload coverage reports more than once

mitchell-codecov:
Coverage reports are currently sent twice:
- [`uploadToCodecov`](https://github.com/codecov/uploader/blob/037e1e74331080fdad25e550d8fdb7785c17b330/src/helpers/web.ts#L94)
- [`uploadToCodecovPUT`](https://github.com/codecov/uploader/blob/037e1e74331080fdad25e550d8fdb7785c17b330/src/helpers/web.ts#L59)

### DIFF
--- a/src/helpers/web.ts
+++ b/src/helpers/web.ts
@@ -91,9 +91,6 @@ export async function uploadToCodecov(
       )}&token=${token}&${query}`,
     )
     .retry()
-    .send(uploadFile)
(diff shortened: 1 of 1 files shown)


drazisil-codecov (approved):
Approved. Please verify that this commit shows correctly in Codecov with no errors and all uploads before merging.

Issue #252, opened 2021-07-24
Unclear how to resolve "Unable to detect service, please specify manually."


slosd:
**Describe the bug**
It's unclear how to resolve "Unable to detect service, please specify manually."
What do I specify manually? Looking at the source here specifying branch etc. manually won't avoid this error.

**To Reproduce**
1. Run uploader on unrecognized provider
2. Receive the error message
3. Unsure how to proceed from here.

**Expected behavior**
Helpful error message which mentions cli arguments or other details that help to make this work on an unrecognized provider. (https://docs.codecov.com/docs/detecting-ci-services doesn't seem to help either)

Alternatively don't fail at all and implement what's suggested here: https://github.com/codecov/uploader/issues/23

**Additional context**
Build: https://uploader.codecov.io/v0.1.0_6352/linux/codecov
Ran the uploder with `env CI= ./codecov -t $TOKEN -s coverage` to force the local provider and that works.

thomasrockhu:
@slosd are you running in a Docker container? Can you also provide a little more of the Codecov output and what CI you are running (or just local)?

slosd:
`env CI= ./codecov -t $TOKEN -s coverage` works, sorry, that failed earlier because of misconfiguration.

So my bug report concerns the fact that setting CI to an empty string is far from obvious given the error message or documentation. I discovered this by looking at https://github.com/codecov/uploader/blob/master/src/ci_providers/provider_local.js. 

I think it would be reasonable to fall back to the "Local" provider automatically. Judging from the issue I linked, it seems to me this actually is the intended behavior.

Full log:
>   Codecov report uploader 0.1.0
> => Project root located at: /home/builder/clone
> => Found 4 possible coverage files:
> coverage-final.json
> clover.xml
> coverage-final.json
> lcov.info
> Error!: Error: Unable to detect service, please specify manually.

running on a linux machine provided by https://codemagic.io

thomasrockhu:
@slosd sorry, you shouldn't have to do that. What CI are you running in and are you running in a container?

slosd:
I'm using https://codemagic.io and the build runs in a (linux) VM. The environment initially has set `CI=true` which prevents the local provider from being selected.

slosd:
Maybe worth mentioning: I still use the bash uploader in another project on the same provider just with `bash codecov` and it has no problems gathering all necessary info.

thomasrockhu:
@slosd, can you share the codecov output of the project that works?

[...]

Issue #23, opened 2020-05-08
Use local provider when matching CI one is not found


drazisil:
**Describe the bug**
Use local provider when matching CI one is not found

**To Reproduce**
1. Try running on a CI provider we don't support yet

**Expected behavior**
Do not fail on undetected CI provider.

drazisil:
@eddiemoore Can you confirm this is the current behavior?

eddiemoore:
@drazisil I believe it is yes

tagatac:
This does not appear to be resolved. See #252 and #284

drazisil:
@drazisil-codecov to reopen and investigate

drazisil:
https://github.com/codecov/uploader/pull/299#issuecomment-898737295

drazisil-codecov:
@tagatac When this PR gets merged into the release branch, it will be available at uoploader.codecov.io

tagatac:
There was one more bug in the web helper (#301), but it looks good now! https://app.travis-ci.com/github/tagatac/bagoup/builds/236032777

PR #299, opened 2021-08-12
Improve error handling and message flexibility 

drazisil-codecov:
This was intended to address #252 which occurs when https://github.com/codecov/uploader/blob/master/src/ci_providers/provider_local.ts is triggered in an environment where `git` does not exist, such as Docker.

In this case you end up the unhelpful message of `Unable to detect service, please specify manually`, which does not explain _how_ to do this .

However, given the refactor needed to get the code to a state where I was ready to address that, I decided to end here and improve the documentation in a fresh PR.


src/helpers/cli.ts
```diff
@@ -0,0 +1,147 @@
+import yargs from 'yargs'
```
mitchell-codecov, line 1:
```suggestion
```

`yargs` appears to be unused in this file.

mitchell-codecov (dismissed):

test/providers/index.test.ts
```diff
@@ -1,13 +1,14 @@
...
 import providers from '../../src/ci_providers'
+import { UploaderInputs } from '../../src/types'
 
 describe('CI Providers', () => {
   afterEach(() => {
     td.reset()
   })
 
-  it('is an array of CI providers', () =>
+  it('check that each provider', () =>
```
mitchell-codecov, line 11: Nested `it` blocks might be somewhat fragile. I don't believe it's an officially supported pattern.

1. Change the `it` on line 11 to `describe`:
   ```suggestion
     describe('check that each provider', () =>
   ```
2. Move the contents of line 12 into its own `it` block

drazisil-codecov: I'm not nesting `it`, am I? 👀

foo.txt
```diff
@@ -0,0 +1,14 @@
...
+   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
+   | |___| (_) | (_| |  __/ (_| (_) \ V /
+    \_____\___/ \__,_|\___|\___\___/ \_/
+
+  Codecov report uploader 0.1.0
+[2021-08-13T17:36:22.168Z] ['info'] => Project root located at: /Users/drazisil-codecov/Downloads
+[2021-08-13T17:36:22.168Z] ['info'] -> No token specified or token is empty
+[2021-08-13T17:36:22.168Z] ['verbose'] Start of network processing...
+[2021-08-13T17:36:22.168Z] ['verbose'] Searching for files in /Users/drazisil-codecov/Downloads
+[2021-08-13T17:36:22.178Z] ['info'] Searching for coverage files...
```
mitchell-codecov, line 14: Make sure to untrack this.

mitchell-codecov (changes requested):

foo.txt
drazisil-codecov: 🤦

drazisil:
@drazisil-codecov #23 is the issue you need to address, where the uploader is ran in an environment where `CI=true`, preventing local from detecting, but no other provider matches.

Best way to handle is probably to have local be outside of the detection chain and the order to be like so:
* check if we have a complete set of manual overrides (slug, SHA)
* if not, loop though all providers (except local)
* If not matched, but `CI=true`, try local with the values for name set to `Generic CI`
* if fails, run local normally
* if fails, display message explaining failure, and explaining that SHA and slug need to be set as args

drazisil:
> @drazisil, why the name change from `envs` to `environment`? Shouldn't we just change all instance of `envs` to `environment` instead?

Yargs has a method called env, so typescript was having a panic attack.

Regarding the change of all instances, I believe I did, except in a couple cases where it's aliased back to env.

mitchell-codecov (approved):

PR #305, opened 2021-08-16
(DO NOT MERGE) Send `token` via HTTP header

mitchell-codecov:


drazisil-codecov (changes requested):
No. Remember, we can't remove this for self hosted.

PR #306, opened 2021-08-16
Remove filename validation

### LINKED ISSUE #303: the `-f` option seems to have stopped working in build `7865` (works in `6943`)
**Describe the bug**
My scheduled CI from 6 days ago succeeded: https://github.com/ansible-collections/community.hashi_vault/runs/3281498133?check_suite_focus=true#step:6:561
The scheduled CI from the next day, with no changes, failed: https://github.com/ansible-collections/community.hashi_vault/runs/3292087124?check_suite_focus=true#step:6:29

I am directly providing every file to be uploaded via `-f` options. This was working fine, now it can't find any files.

I tried this locally as well, downloading the latest version of the script, and I can't get it to recognize any files that way. Adding verbose, it looks like it's ignoring the `-f` option (though not completely, because I don't add it and it can't find any files, it says I should use `-f` to specify the files to upload).

**To Reproduce**
```
codecov -f path/to/file.xml
```

**Expected behavior**
Specified file(s) uploaded

**Screenshots**
image

**Additional context**
The thing that puzzled me is that both before and after it broke, the version returned by the binary is `0.1.0`. Looking at your releases page it now seems like you release new versions without changing the version number? If so, that means I don't have any way of trying to download an older version to see if the behavior really changed. I also can't get the binary to show me a specific build/commit which would have also helped diagnose the difference.

A new build was released right around when the behavior seemingly changed. It contains 16 commits but I can't really tell which if any of them may have contributed to it:
https://github.com/codecov/uploader/compare/v0.1.0_7865...master

Edit: I see that there are tests for finding specific files with `-f`, and presumably those tests hadn't failed, so I'm not entirely sure what I'm seeing, I just know it was working 6+ days ago, and since then, it isn't :(

---

### confirmed
I put up a PR in my repo ( https://github.com/ansible-collections/community.hashi_vault/pull/129 ) that downloads the previous build direct from the github release, and with that previous release everything is working: https://github.com/ansible-collections/community.hashi_vault/runs/3334647366?check_suite_focus=true#step:6:362

I also confirmed this locally.

So, build 6943 works fine.
The build I get when downloading "latest" via https://uploader.codecov.io/latest/linux/codecov (which I presume is build 7865) is not working.

mitchell-codecov:
The current filename validation is too strict; furthermore, it is not currently necessary since the underlying file system is capable of raising an error if it receives an invalid file name.

Fixes #303

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -149,8 +149,7 @@ export async function main(
 
   // Remove invalid and duplicate file paths
   coverageFilePaths = [... new Set(coverageFilePaths.filter(file => {
-    return validateHelpers.validateFileNamePath(file) &&
-      fileExists(args.dir || projectRoot, file)
+    return fileExists(args.dir || projectRoot, file)
   }))]
 
   if (coverageFilePaths.length > 0) {


drazisil-codecov (approved):
Looks good to me. This keeps us from dropping directories , correct?

briantist:
thanks for the quick fix! do you have an expected build / version number for this?

Also am concerned that no tests caught this and there don't seem to be any added here. I'll be pinning versions/builds from now on but would be great to see some tests to catch regressions; I'd like to eventually be confident enough to pull latest again.

PR #308, opened 2021-08-16
fix: Parse multiple flags properly

thomasrockhu-codecov:
Flags are now getting passed in as a `string`, not a `string[]`, so the prior validation would throw out the entire string.

### DIFF
--- a/src/helpers/web.ts
+++ b/src/helpers/web.ts
@@ -21,15 +21,17 @@ export function populateBuildParams(
   const { args, environment: envs } = inputs
   serviceParams.name = args.name || envs.CODECOV_NAME || ''
   serviceParams.tag = args.tag || ''
+
   let flags: string[]
   if (typeof args.flags === 'object') {
     flags = [...args.flags]
   } else {
-    flags = [args.flags || '']
+    flags = (args.flags || '').split(',')
   }
   serviceParams.flags = flags
     .filter(flag => validateHelpers.validateFlags(flag))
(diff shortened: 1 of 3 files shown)


src/helpers/web.ts
```diff
@@ -21,15 +21,12 @@ export function populateBuildParams(
...
   serviceParams.name = args.name || envs.CODECOV_NAME || ''
   serviceParams.tag = args.tag || ''
-  let flags: string[]
-  if (typeof args.flags === 'object') {
-    flags = [...args.flags]
-  } else {
-    flags = [args.flags || '']
-  }
+
+  const flags = (args.flags || '').split(',')
```
mitchell-codecov, line 25:
```ts
    let flags: string[]
    if (typeof args.flags === 'object') {
      flags = [...args.flags]
    } else {
      flags = (args.flags || '').split(',')
    }
```

src/types.ts
```diff
@@ -5,7 +5,7 @@ export interface UploaderArgs {
   env?: string // Specify environment variables to be included with this build
   sha?: string // Specify the commit SHA mannually
   file?: string | string[] // Target file(s) to upload
-  flags?: string | string[] // Flag the upload to group coverage metrics
+  flags?: string // Flag the upload to group coverage metrics
```
drazisil-codecov, line 8: How is the splitting and handling of flags different from files?

drazisil-codecov (approved):

PR #321, opened 2021-08-19
fix: trim git output

drazisil-codecov:


mitchell-codecov (approved):

test/providers/provider_bitbucket.test.ts
```diff
@@ -158,9 +158,9 @@ describe('Bitbucket Params', () => {
       slug: 'testOwner/testSlug',
     }
 
-    const execFileSync = td.replace(childProcess, 'execFileSync')
+    const execFileSync = td.replace(childProcess, 'spawnSync')
     td.when(execFileSync('git', ['rev-parse', 'testingsha12'])).thenReturn(
-      'newtestsha',
+      { stdout: 'newtestsha'},
```
vlad-ko, line 163: Hi, should we a test here for the SHA with \n character (s) or other whitespace around the SHA, to confirm that the problem we're solving for is fixed?

PR #342, opened 2021-08-27
Added CI Script to Evaluate Uploader's Report Coverage Accuracy at PR Time

adrian-codecov:
This PR is to create a script that runs in CI to see if any changes in Uploader change the way we calculate code coverage. This is done by comparing code coverage produced during CI with Production known coverage, using each Standard's known coverage as reference (see [here](https://github.com/codecov/standards/tree/master/coverage_data)). The script will alert at PR time if there is a discrepancy during the comparison. 

This whole process is done by spinning Codecov's infrastructure using standardized containers which contains the comparison script. This script is meant to catch any unforeseen behavior during CI when computing code coverage.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -36,7 +36,7 @@
         "eslint": "7.32.0",
         "eslint-config-prettier": "8.3.0",
         "husky": "7.0.2",
-        "jest": "27.0.6",
+        "jest": "27.1.0",
         "jest-spec-reporter": "1.0.17",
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -468,4 +488,4 @@ workflows:
             - deploy
           filters:
             branches:
-              only: release
+              only: release
```
mitchell-codecov, line 491: This should end in a newline.

PR #358, opened 2021-09-13
Added CI Script to Evaluate Uploader's Report Coverage Accuracy at PR Time

adrian-codecov:
This PR is to create a script that runs in CI to see if any changes in Uploader change the way we calculate code coverage. This is done by running a script that compares code coverage calculated during CI with Production known coverage, using each Standard's known coverage as reference. The script will alert at PR time if there is a discrepancy during the comparison.


Makefile
```diff
@@ -28,3 +28,7 @@ build_alpine:
 	npm run build-alpine
 
 .PHONY: clean install test build
+
+run-standards-comparison:
```
mitchell-codecov, line 32:
```suggestion
.PHONY: run-standards-comparison
run-standards-comparison:
```

Make sure to add the target of each synthetic/phony recipe as a `.PHONY` target. This allows Make to note that a file by the name of "run-standards-comparison" is not the actual target of the recipe but is rather its name.

Note that each `.PHONY` definition denotes an additional dependency of `.PHONY` rather than a reassignment such that any previous recipes are overridden. In other words, `.PHONY` should have a dependence on each synthetic rule. This can be done all at once (similar to the recipe on [L30](https://github.com/codecov/uploader/blob/6843f5acd933184a1de773cc1a32ad4519785e6e/Makefile#L30)) or on an individual basis. The benefit of creating an individual `.PHONY` recipe for each phony target is that they can be written in a close proximity to each other which would allow for easier maintenance.

mitchell-codecov: To illustrate how "doing it all at once" may be a detriment, note how several of the phony targets are missing from the `.PHONY` recipe.

adrian-codecov: Awesome, good suggestion! Just read up a bit on makefiles and agree with your change, thanks for that @mitchell-codecov!

mitchell-codecov (approved):

drazisil-codecov (approved):

PR #363, opened 2021-09-15
Encode query strings

### LINKED ISSUE #255: `generateQuery` does not encode query strings
Use the native [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) API as an abstraction for handling query strings. It both formats and encodes/decodes query strings.

The current tests for `generateQuery` appear to validate the return values against malformed URIs (e.g., slashes and whatnot are not encoded properly). In other words, any refactor to `generateQuery` will have to correct the tests.

mitchell-codecov:
Fixes #255


drazisil-codecov (approved):
Looks good, assuming the uploads pass.

PR #367, opened 2021-09-16
Fix encoding of `buildURL`

mitchell-codecov:


drazisil-codecov (changes requested):
Please adjust
* Unit tests to testing that encoding works (do not start encoded)
* Integration (upload) tests to make sure we pass _all_ query params so we we can catch if server break before we deploy.

PR #375, opened 2021-09-20
Always use CI detection

mitchell-codecov:


src/ci_providers/provider_githubactions.ts
```diff
@@ -66,7 +66,7 @@ function _getSHA(inputs: UploaderInputs): string {
   const pr = _getPR(inputs)
 
   let commit = envs.GITHUB_SHA
-  if (pr && pr !== 0 && !args.sha) {
+  if (pr && pr !== '0' && !args.sha) {
```
drazisil-codecov, line 69: I think this is part of a PR detection process down the road.

https://github.com/codecov/uploader/blob/7ba8b19bd516d7bcb8cccdae75a178323870d238/src/helpers/web.ts#L114

Since a PR can not be 0 (not a number), it gets converted to ''

drazisil-codecov (changes requested):

src/ci_providers/provider_githubactions.ts
mitchell-codecov, line 69:
```suggestion
  if (pr && pr !== '' && !args.sha) {
```

drazisil-codecov (approved):

PR #379, opened 2021-09-22
Standard changelog

### LINKED ISSUE #311: Include build number as part of version or in program output
Related: #303 

Posting this as a new request so it doesn't get lost, since #303 was already closed by the PR fix to the underlying issue (thank you!)

As is, version `0.1.0` (for example) does not refer to a specific release of the uploader; [there are many builds](https://github.com/codecov/uploader/releases) that identify as `0.1.0`. In addition, `codecov --version` does not report the additional build number.

The main issues with this are:
- a version doesn't correspond to a specific set of code/functionality
- it's difficult to determine which actual build of the binary is in use (for troubleshooting)
- it violates the principle of least surprise

### Describe the solution you'd like

In order of preference:
- the version number should be incremented with each build (`0.1.1`, `0.1.2`, etc.)
- the build can become part of the version (`0.1.0.8278`, `0.1.0.7865`, etc.)
- the build could remain separate but be reported from the program output (`0.1.0 build 7865`)

### Describe alternatives you've considered
The only thing I can do to prevent the negative side effects is pin versions (there are no instructions I could find that show the version format in your URLs, but I was able to find it by browsing the paths).

### Additional context
In #303, the only way I could confirm the issue was to test each one side by side; I could not just look at the output from my CI to see clearly that the build changed along with the behavior change.

### LINKED ISSUE #327: Add command to display changelog
Should help somewhat with #311

drazisil-codecov:
Closes #311 
Closes #327


mitchell-codecov (approved):

briantist:
@drazisil-codecov it looks like there should be a 0.1.1 and/or 0.1.2 release based on this? And that maybe future builds will be incrementing the version? is that correct? It's now in the CHANGELOG but they don't appear in the GitHub releases nor in the downloads ( https://uploader.codecov.io/linux ). When will they appear?

PR #389, opened 2021-09-28
feat: add support to esclude search path

drazisil-codecov:


src/index.ts
```diff
@@ -149,14 +152,38 @@ export async function main(
...
-    args.dir || projectRoot,
-    coverageFilePaths.length > 0 ? coverageFilePaths : coverageFilePatterns(),
-  ))
+
+let searchPath = projectRoot
+
+// If customer provided an exclude to -s
+if (args.dir) {
+    if (args.dir.startsWith('!') || args.dir.startsWith('^')) {
+      verbose(`Adding search path to ignore list: ${args.dir}`)
```
drazisil-codecov, line 161: When ran with `-s ^test,coverage.\*` it gets added to the ignore list as `test,coverage.*` as I would expect. However, we still locate files named coverage.*`

What am I missing here, @mitchell-codecov ?

PR #390, opened 2021-09-28
Negate paths

mitchell-codecov:


src/index.ts
```diff
@@ -74,7 +74,9 @@ function dryRun(
 export async function main(
   args: UploaderArgs,
 ): Promise<void | Record<string, unknown>> {
-
+  
+  console.debug({ args });
+
```
drazisil-codecov, line 79:
```suggestion
  verbose(`Arguments passed: ${args}`, Boolean(args.verbose))
```

src/index.ts
```diff
@@ -149,9 +151,18 @@ export async function main(
       coverageFilePaths = args.file
     }
   }
+  const isNegated = (path: string) => path.startsWith('!')
```
drazisil-codecov, line 154: Please write a test to confirm this behaves as expected.

drazisil-codecov (changes requested):

src/index.ts
mitchell-codecov: That line was mistakenly added. I've since removed it.

src/index.ts
mitchell-codecov: Since `isNegated` is just a helper function for the anonymous arrow function a bit below, its behavior should be caught by the E2E test I pushed in one of my more recent commits.

src/index.ts
```diff
@@ -149,9 +149,18 @@ export async function main(
...
-    coverageFilePaths.length > 0 ? coverageFilePaths : coverageFilePatterns(),
+    (() => {
+      const numOfNegatedPaths = coverageFilePaths.filter(isNegated).length
+      
+      if (coverageFilePaths.length > numOfNegatedPaths) {
+        return coverageFilePaths
+      } else {
+        return coverageFilePaths.concat(coverageFilePatterns())
+      }
+    })(),
```
mitchell-codecov, line 163: The idea here is that, if the user only specifies exclusion filters, the default coverage file patterns are used.

eliatcodecov (approved):

drazisil-codecov (approved):

PR #483, opened 2021-11-02
Fix false positive detection for Heroku-provider

FredrikSeidl:
This PR fixes an error where Jenkins is incorrectly detected as Heroku, thus breaking the upload of coverage data since the provider cannot find the proper SHA for the report.

The problem is that the environment variable `HEROKU_TEST_RUN_BRANCH` is `undefined` which when losely compared to empty string will become `true` whereas what we really want to know is if it actually has any value. Looked through the other providers and you seem to like checking this by casting it as a boolean so I decided to do the same

image


mitchell-codecov (approved):
Excellent!

FredrikSeidl:
@mitchell-codecov thank you for the merge. When can I expect to see this live? It's currently affecting our builds so either I hang in until you release or change our scripts to use a version prior to the bug was introduced

PR #492, opened 2021-11-05
chore: create .editorconfig

mitchell-codecov:


.prettierrc
```diff
@@ -10,7 +10,5 @@
   "requirePragma": false,
   "semi": false,
   "singleQuote": true,
-  "tabWidth": 2,
```
mitchell-codecov, line 13: This corresponds to the addition of the EditorConfig-managed [`indent_size`](https://github.com/codecov/uploader/blob/0743c8b/.editorconfig#L10).

.prettierrc
```diff
@@ -10,7 +10,5 @@
...
-  "trailingComma": "all",
-  "useTabs": false
```
mitchell-codecov, line 15: This corresponds to the addition of the EditorConfig-managed [`indent_style`](https://github.com/codecov/uploader/blob/0743c8b/.editorconfig#L9).

drazisil-codecov (approved):
~Would this replace the formatter's rules?~

.prettierrc
drazisil-codecov: 🤦‍♀️ I thought that was this file. Too early 😹

.prettierrc
drazisil-codecov: 🤦‍♀️ I thought that was this file. Too early 😹

Davinci5555 (approved):

PR #494, opened 2021-11-08
chore: use code regions

mitchell-codecov:
This PR modifies the sectional comments in index.ts with fold regions. The [Folding](https://code.visualstudio.com/docs/editor/codebasics#_folding) section from the VS Code user guide goes over the feature as it exists in VS Code.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -93,7 +93,7 @@ export async function main(
   Step 7: either upload or dry-run
   */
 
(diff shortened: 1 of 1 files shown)


drazisil-codecov (commented):
What are these?

drazisil-codecov (approved):
Would love some docs in the description.

Issue #378, opened 2021-09-22
[CODE-878] The Uploader should support proxy functionality


eliatcodecov:
**Is your feature request related to a problem? Please describe.**

This issue captures the few discussion items in #219, as it seemed like a clear, and separate, feature request for supporting proxy functionality had emerged.  In short, users require a means to proxy requests made by the uploader in order to support their use cases properly. The bash uploader supported this approach via the `-U` flag which passed additional parameters to the bash uploader's internal curl command. The binary uploader does not use curl internally for requests, so will need to find an alternative means to support this behavior.

**Describe the solution you'd like**
The user should be able to specify an upload proxy via use of a command line argument. 

**Describe alternatives you've considered**
There really are none. Without this functionality some use cases simply cannot be supported.

**Additional context**
The reader is directed to the final comments in #219 for any additional context.

PR #495, opened 2021-11-08
[CODE-878] add proxy functionality

drazisil-codecov:
Solves #378 

This PR introduces the `-U` (upstream) flag, which provides a means to use the uploader with a proxy.

It is hopefully easy to understand, the usage to connect to a proxy of `localhost:3128`, for example, would be `-U https://localhost:3128`

It has not been tested with authentication , please report any issues if that fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   },
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
-    "@types/node-fetch": "2.5.12",
     "fast-glob": "3.2.7",
     "glob": "7.2.0",
     "https-proxy-agent": "5.0.0",
@@ -39,8 +38,6 @@
     "line-reader": "0.4.0",
     "node-fetch": "2.6.6",
     "snake-case": "3.0.4",
-    "superagent": "6.1.0",
-    "typescript": "4.4.4",
     "validator": "13.7.0",
     "yargs": "17.2.1"
   },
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -31,10 +31,13 @@
   },
   "homepage": "https://github.com/codecov/uploader#readme",
   "dependencies": {
+    "@types/node-fetch": "2.5.12",
```
mitchell-codecov, line 34: This should be in `devDependencies`.

src/types.ts
```diff
@@ -21,6 +23,7 @@ export interface UploaderArgs {
   url?: string // Change the upload host (Enterprise use)
   clean?: string // Move discovered coverage reports to the trash
   feature?: string // Toggle features
+  upstream: string // Upstream proxy to connect to
```
mitchell-codecov, line 26: Can we make this optional so we don't have to default it to an empty string everywhere else?

src/types.ts
```diff
@@ -55,3 +58,13 @@ export interface IServiceParams {
...
+
+export interface IRequestHeaders {
+  url: string;
+  options: {
+    method: string,
+    agent?: Agent
+    body?: string | Buffer,
+    headers: Record<string, string>
+  } 
+}
```
mitchell-codecov, line 70: Can this be replaced with an export from the types package?

mitchell-codecov (approved):
Do we actually look for codecov.yml in the .github folder?

src/types.ts
drazisil-codecov: No, it's a custom object.

package.json
drazisil-codecov: > Do we actually look for codecov.yml in the .github folder?

Yes, see https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml

Issue #529, opened 2021-12-02
Blacklist not working - Scripts included in uploads


Flamefire:
**Describe the bug**
The uploader is supposed to filter and NOT upload e.g `*.sh` and `*.ps1` files but yet it does as evident by the following log:

```
['verbose'] Preparing to clean the following coverage paths: coverage.info,ci/codecov.ps1,ci/codecov.sh,ci/github/codecov.sh,ci/travis/codecov.sh
['info'] => Found 5 possible coverage files:
  coverage.info
  ci/codecov.ps1
  ci/codecov.sh
  ci/github/codecov.sh
  ci/travis/codecov.sh
```

**Expected behavior**
Only uploading coverage files, not shell scripts.
A way to disable the glob search.

**Additional context**
See also #528

Flamefire:
I proved this behavior with tests and added a PR to fix this: #531

drazisil-codecov:
I believe this is resolved

PR #531, opened 2021-12-03
Fix: Don't collect scripts as coverage files

Flamefire:
- Add tests to check that scripts and config files are removed from coverage files by default: I expect this to fail, see #529
- Add the missing subdir globstar

### DIFF
--- a/test/helpers/files.test.ts
+++ b/test/helpers/files.test.ts
@@ -294,7 +294,8 @@ describe('File Helpers', () => {
     })
 
     it("returns the input array when passed an empty ignore array", async() => {
-      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), await paths, [])).toEqual(paths)
+      const pathsSync = await paths
+      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), pathsSync, [])).toEqual(pathsSync)
     })
 
     it("ignores an ignore filename", async() => {
--- a/test/helpers/files.test.ts
+++ b/test/helpers/files.test.ts
@@ -228,13 +228,13 @@ describe('File Helpers', () => {
 
     it('can return a list of coverage files with a pattern', async () => {
       expect(
-        await fileHelpers.getCoverageFiles('.', ['index.test.ts']),
-      ).toStrictEqual(['test/index.test.ts', 'test/providers/index.test.ts'])
+        await fileHelpers.getCoverageFiles('.', ['coverage.txt']),
+      ).toStrictEqual(['test/fixtures/coverage.txt', 'test/fixtures/other/coverage.txt'])
     })
     it('can return a list of coverage files with a negated pattern', async () => {
       expect(
-        await fileHelpers.getCoverageFiles('.', ['index.test.ts', '!test/providers']),
-      ).toStrictEqual(['test/index.test.ts'])
+        await fileHelpers.getCoverageFiles('.', ['coverage.txt', '!test/fixtures/other']),
+      ).toStrictEqual(['test/fixtures/coverage.txt'])
     })
     describe('coverage file patterns', () => {
       it('contains `jacoco*.xml`', () => {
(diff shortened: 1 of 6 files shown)


test/helpers/files.test.ts
```diff
@@ -282,47 +282,43 @@ describe('File Helpers', () => {
   })
 
   describe("cleanCoverageFilePaths()", () => {
+    const ignoreGlobs = fileHelpers.getBlocklist()
+    const paths = fileHelpers.getCoverageFiles(
+      '.',
+      fileHelpers.coverageFilePatterns(),
+    )
```
mitchell-codecov, line 289: I see what you're doing here. Instead of directly invoking the shared logic here, however, modify `paths` to return the invocation such that the individual tests are still the ones calling `fileHelpers.getCoverageFiles`:

```suggestion
    const paths = () => fileHelpers.getCoverageFiles(
      '.',
      fileHelpers.coverageFilePatterns(),
    )
```

With the above suggestion, `paths` changes ever so slightly. It may make sense to update its name accordingly. `getPaths` makes sense here.

As an aside, it's clear here that `fileHelpers.getCoverageFiles` is not very ergonomic for the (very common) use case of _getting coverage files for the current directory_. Perhaps a future PR can update it to use better defaults.

test/helpers/files.test.ts
```diff
@@ -282,47 +282,43 @@ describe('File Helpers', () => {
   })
 
   describe("cleanCoverageFilePaths()", () => {
+    const ignoreGlobs = fileHelpers.getBlocklist()
```
mitchell-codecov, line 285: Perhaps it makes sense for the tests to contain this logic directly; e.g.:

```ts
it("works", async () => {
  // …
  expect(
    () => cleanCoverageFilePaths(cwd(), paths, getBlocklist())
  ).not.toThrow();
});
```

test/helpers/files.test.ts
```diff
@@ -282,47 +282,43 @@ describe('File Helpers', () => {
...
     it("ignores an ignore filename globstar", async() => {
-      const paths = await fileHelpers.getCoverageFiles(
-        '.',
-        fileHelpers.coverageFilePatterns(),
-      )
-      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), paths, ["**/other/*"])).not.toContainEqual(expect.stringMatching('other'))
+      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), await paths, ["**/other/*"])).not.toContainEqual(expect.stringMatching('other'))
+    })
+    
+    it("ignores shell scripts by default", async() => {
```
mitchell-codecov, line 314:
```suggestion
    it("ignores shell scripts by default", async () => {
```

test/helpers/files.test.ts
```diff
@@ -282,47 +282,43 @@ describe('File Helpers', () => {
...
+      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), await paths, ignoreGlobs)).not.toContainEqual(expect.stringMatching('codecov.sh'))
+    })
+    it("ignores powershell scripts by default", async() => {
```
mitchell-codecov, line 317:
```suggestion
    it("ignores powershell scripts by default", async () => {
```

test/helpers/files.test.ts
```diff
@@ -282,47 +282,43 @@ describe('File Helpers', () => {
...
+      expect(fileHelpers.cleanCoverageFilePaths(process.cwd(), await paths, ignoreGlobs)).not.toContainEqual(expect.stringMatching('codecov.ps1'))
+    })
+    it("ignores codecov configs by default", async() => {
```
mitchell-codecov, line 320:
```suggestion
    it("ignores codecov configs by default", async () => {
```

mitchell-codecov (changes requested):
Thanks for your work on this PR.

mitchell-codecov (approved):

drazisil-codecov (approved):

PR #532, opened 2021-12-03
fix: alternative env vars for drone build url

Flamefire:
See e.g. https://0-8-0.docs.drone.io/environment-reference/

### DIFF
--- a/src/ci_providers/provider_drone.ts
+++ b/src/ci_providers/provider_drone.ts
@@ -14,7 +14,7 @@ function _getBuild(inputs: UploaderInputs): string {
 
 function _getBuildURL(inputs: UploaderInputs): string {
   const { environment: envs } = inputs
-  return envs.DRONE_BUILD_URL || ''
+  return envs.DRONE_BUILD_LINK || envs.DRONE_BUILD_URL || envs.CI_BUILD_URL || ''
 }
 
 function _getBranch(inputs: UploaderInputs): string {
--- a/test/providers/provider_drone.test.ts
+++ b/test/providers/provider_drone.test.ts
@@ -42,6 +42,34 @@ describe('Drone Params', () => {
         DRONE_COMMIT_SHA: 'testingsha',
         DRONE_BUILD_NUMBER: '2',
         DRONE_PULL_REQUEST: '1',
+        DRONE_BUILD_LINK: 'https://www.drone.io/',
+        DRONE_REPO: 'testOrg/testRepo',
+      },
+    }
+    const expected: IServiceParams = {
+      branch: 'master',
+      build: '2',
+      buildURL: 'https://www.drone.io/',
+      commit: 'testingsha',
+      job: '',
+      pr: '1',
+      service: 'drone.io',
+      slug: 'testOrg/testRepo',
+    }
+    const params = providerDrone.getServiceParams(inputs)
+    expect(params).toMatchObject(expected)
+  })
+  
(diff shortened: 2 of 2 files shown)


src/ci_providers/provider_drone.ts
```diff
@@ -14,7 +14,7 @@ function _getBuild(inputs: UploaderInputs): string {
 
 function _getBuildURL(inputs: UploaderInputs): string {
   const { environment: envs } = inputs
-  return envs.DRONE_BUILD_URL || ''
+  return envs.DRONE_BUILD_LINK || envs.DRONE_BUILD_URL || envs.CI_BUILD_URL || ''
```
mitchell-codecov, line 17: I can't find any documentation for `CI_BUILD_URL`. Is it documented somewhere?

mitchell-codecov (changes requested):
`DRONE_BUILD_LINK` should be integrated into a test.

src/ci_providers/provider_drone.ts
Flamefire: Found that in other sources collecting coverage for drone, not sure anymore where
Google turns up this e.g. https://codeclimate.com/github/codeclimate/ruby-test-reporter/lib/code_climate/test_reporter/ci.rb/source
It seems like a generic variable, so maybe we can keep that as a fallback?

Flamefire:
What's going on on CI?

> 2021-12-04 10:29:25 ERROR 503: Egress is over the account limit..

src/ci_providers/provider_drone.ts
drazisil-codecov: This appears to no longer(?) be a valid DroneCI env https://github.com/drone/runner-go/blob/f7a8ed14f07fb159568f28cd3e2841ce90417744/environ/environ.go

Flamefire: Might be. But it seems it was and as it comes last it should be a better fallback than an empty string, isn't it?

drazisil-codecov (approved):

sdarwin:
Hi @Flamefire you mentioned  https://0-8-0.docs.drone.io/environment-reference/ in the pull request description.  Notice 0-8-0 in the url that's a previous version of drone. The newest variables may be found elsewhere such as https://docs.drone.io/pipeline/environment/reference/

Flamefire:
@sdarwin Are there new variables not yet covered? The uploader should understand both I guess

sdarwin:
The change in this PR is probably ok.   
DRONE_BUILD_LINK is mentioned in the 0.8.0 docs.  
DRONE_BUILD_URL is even older, from 8 years ago https://github.com/drone/docs.drone.io/blob/master/env.markdown
The newest docs don't mention either.   
A google result:
```
drone/drone - Gitterhttps://gitter.im › drone › drone
Jun 15, 2020 — It looks like DRONE_BUILD_LINK environment variable is missing from the docs, or I am blind. Arve Knudsen. @aknuds1.
```

Flamefire:
So the latest drone has NO env var for to the link for the current build? For the one running at https://drone.cpp.al/:  `DRONE_BUILD_URL: undefined`

sdarwin:
In the newest drone, testing shows:
DRONE_BUILD_URL: undefined
DRONE_BUILD_LINK : correctly set

Since this pull request adds DRONE_BUILD_LINK it's probably the right thing to do.

Issue #525, opened 2021-11-30
Misdetection of latest HEAD in PRs on Github Action CI


Flamefire:
**Describe the bug**
I have a warning in the codecov comment on Github after running CI on Github Actions:

>  Current head `<SHA-merge>` differs from pull request most recent head `<SHA-branch>`. Consider uploading reports for the commit `<SHA-branch>` to get more accurate results

The mentioned SHAsums are those of the merge-commit-to-be and the latest commit on the branch.
I'm not sure if that is a bug in the uploader or the reporting later on so pass this on to the correct team if wrong.

**To Reproduce**
1. Go to https://github.com/boostorg/boost-ci/pull/114#issuecomment-982569664
1. See error

**Expected behavior**
No warning. It makes sense to that coverage for the merge-commit is uploaded, not that from the branch.

**Additional context**
The uploader is used "vanilla", i.e. 
```
    curl -Os https://uploader.codecov.io/latest/linux/codecov
    chmod +x codecov
    ./codecov -f coverage.info
```

However I see that on Github actions the commit mentioned in the upload URL (log output) is the merge-SHA while on Drone CI it is the branch-SHA. So I'd say either one is wrong?

drazisil-codecov:
Thanks for the report and research , @Flamefire 

We'll see what we can do to resolve this. As a workaround, you should be able to pass the correct value to `override_pr` as an action key, or `-P` as a uploader flag.

PRs welcome! :)

drazisil-codecov:
ah, I misunderstood. In this case, I believe you want to only upload coverage. for pushes, not pull_request.

https://github.com/boostorg/boost-ci/commit/20f52fe0905ca896125cf3be77a3af573559411d sats it's not part of any branch, which usually means it's either been squashed or otherwise deleted. This can cause issues, but it looks like it's showing correctly on Codecovm so it might be a caching issue with updating the PR. How long ago did it complete?

https://app.codecov.io/gh/boostorg/boost-ci/compare/114/commits

Flamefire:
> I believe you want to only upload coverage. for pushes, not pull_request.

Erm, no? Of course I want to have coverage info for PRs as that is part of CI testing: That a change does have decent coverage which is a condition for the merge.

> boostorg/boost-ci@20f52fe sats it's not part of any branch

Just check the link above or the whole PR: https://github.com/boostorg/boost-ci/pull/114
[...]

PR #533, opened 2021-12-03
Enhance logging of Github Action merge commit message

Flamefire:
Add logging information for the fix of merge commits or when it was not done/required

See #525 for context


src/ci_providers/provider_githubactions.ts
```diff
@@ -4,7 +4,7 @@
 import { IServiceParams, UploaderEnvs, UploaderInputs } from '../types'
 
 import childProcess from 'child_process'
-import { info } from '../helpers/logger'
+import { info, UploadLogger } from '../helpers/logger'
```
drazisil-codecov, line 7:
```suggestion
import { info, verbose } from '../helpers/logger'
```

src/ci_providers/provider_githubactions.ts
```diff
@@ -66,15 +66,18 @@ export function getServiceName(): string {
...
 
   let commit = envs.GITHUB_SHA
-  if (pr && pr !== '' && !args.sha) {
+  if (pr) {
     const mergeCommitRegex = /^[a-z0-9]{40} [a-z0-9]{40}$/
     const mergeCommitMessage = childProcess
       .spawnSync('git', ['show', '--no-patch', '--format="%P"'])
       .stdout.toString()
       .trimRight()
+    UploadLogger.verbose(`Handling PR with parent hash(es) '${mergeCommitMessage}' of current commit.`)
```
drazisil-codecov, line 80:
```suggestion
    verbose(`Handling PR with parent hash(es) '${mergeCommitMessage}' of current commit.`)
```

drazisil-codecov (approved):
Minor change with the logger import, otherwise good! 👍

Flamefire:
Alright, as this is required for  #525 can you tell me when this changeset occurs in the release?

Flamefire:
I've just seen that your change https://github.com/codecov/uploader/pull/533/commits/7d850089bf80eba6629613e84b3a49f11c285179 basically broke the functionality as the `verbose` function requires a second parameter. IMO that function is badly designed anyway and should simply call the `UploadLogger` function instead of requiring callers to pass a second parameter which is already set in https://github.com/codecov/uploader/blob/master/src/helpers/logger.ts#L59

So all I see is: `Commit with SHA 2f652c75e02f52915c45584ab3ee90a75694c78e of PR 125 is not a merge commit` which refers to the merge commit https://github.com/boostorg/boost-ci/commit/2f652c75e02f52915c45584ab3ee90a75694c78e, see https://github.com/boostorg/boost-ci/runs/4653741028?check_suite_focus=true
What I don't see is the reason for that -.-

Flamefire:
I added PR #557 which more or less reworks the logging system. The mix of free and class methods is unfortunate and I more or less removed that using UploadLogger in all cases possible (for `verbose`)

Flamefire:
> It looks like a merge commit

Because it is as shown by Github ;)

> You have uploaded merge commit coverage in the past ...

Yes, by working around the bug in that commit which should get reverted: `export GITHUB_SHA=$(git show --no-patch --format="%P" | awk '{print $NF}')`. So basically yes, it is what we expect. Question is now only: What does the uploader see? This is what the added logging is for. I'll add a temporary commit printing this for you to verify, but well, we need to have that logging to see what is actually wrong

> On the server side Codecov did exactly as you'd expect:

Interesting. So the codecov server does some additional processing that should actually be done by the uploader? Well at least it is not consistent as seen in #525 where the bot posts wrong comments using the merge SHA

Flamefire:
@drazisil-codecov Got the output as requested (the part in single quotes): `Commit sha: 'd9667e22fea3b7931c92ac1602aba8e0b48afea3 dcddcce2c07efcd3bdbf08ded383bf5f8f3bf8d8'`, see https://github.com/boostorg/boost-ci/runs/4668882220?check_suite_focus=true

As mentioned: From there it is (still) correct, but something in the uploader is odd.

Flamefire:
That's why I added the quotes. No line break and it seems to be indeed a space. C&P that to e.g. notepad++ matches the regex used (`^[a-z0-9]{40} [a-z0-9]{40}$`). That's why we need the log info from https://github.com/codecov/uploader/pull/533/files#diff-c41f8d0217d4b3e99bb91e8bae797e2574cd37e1445295b142d294fb6a794aacR77 to tell what the uploader actually gets.

PR #534, opened 2021-12-03
Fix SHA usage on PRs for appveyor

Flamefire:
For PRs the HEAD commit SHA of the source branch is in `$APPVEYOR_PULL_REQUEST_HEAD_COMMIT` so use that if it is set.

### DIFF
--- a/src/ci_providers/provider_appveyorci.ts
+++ b/src/ci_providers/provider_appveyorci.ts
@@ -58,7 +58,7 @@ export function getServiceName(): string {
 
 function _getSHA(inputs: UploaderInputs) {
   const { args, environment: envs } = inputs
-  return args.sha || envs.APPVEYOR_REPO_COMMIT || ''
+  return args.sha || envs.APPVEYOR_PULL_REQUEST_HEAD_COMMIT || envs.APPVEYOR_REPO_COMMIT || ''
 }
 
 function _getSlug(inputs: UploaderInputs) {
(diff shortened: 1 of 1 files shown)


test/providers/provider_appveyorci.test.ts
```diff
@@ -59,14 +59,46 @@ describe('AppveyorCI Params', () => {
         APPVEYOR_BUILD_VERSION: '3',
         APPVEYOR_JOB_ID: '1',
         APPVEYOR_PROJECT_SLUG: 'testRepo',
-        APPVEYOR_PULL_REQUEST_NUMBER: '4',
```
mitchell-codecov, line 62: Is there a specific reason why this line was removed?

Flamefire: Yes. For PRs appveyor sets e.g. `$APPVEYOR_PULL_REQUEST_HEAD_COMMIT`
So a PR run has to be tested differently with different outcome. With the (mock) variables in this test this isn't a PR build.
I basically copied this test and added the missing env var: https://github.com/codecov/uploader/pull/534/files#diff-14c6e685ddd586e988a9b18bfb02307dd44a512dc8615e90eb2bb56b2b7b35d5R83

Flamefire:
@mitchell-codecov Is there any documentation about the expected values for `commit` and `branch`? This change is based on the observation that the github actions provider code does some serious effort to extract the source branch head SHA and hence that is the required value.
For `branch` I'm not sure...

See also https://github.com/codecov/uploader/issues/525#issuecomment-985726111

test/providers/provider_appveyorci.test.ts
drazisil-codecov: I don't use Appveyer myself. 🤔 

@blueyed, as a fellow Appveyor user, do you mind chiming in here? You helped a lot with the bash uploader's Appveyor support if I recall.

Flamefire:
As per your comment at https://github.com/codecov/uploader/issues/525#issuecomment-995783190 this change is correct and can also be verified by checking the appveyor docs: https://www.appveyor.com/docs/environment-variables/

drazisil-codecov (approved):
Going to merge, appears correct.

PR #535, opened 2021-12-06
chore: swap spawnSync calls for helper function runExternalProgram

michael-codecov:
Note: the specified encoding should be redundant since we call toString on the buffer response in `runExternalProgram`

https://github.com/codecov/uploader/blob/722c8191c20cda5fba026f889069ed40ef8da4e6/src/helpers/util.ts#L15


mitchell-codecov (approved):
A future PR should create an abstraction for each use case. As an example, the call to `git rev-parse --show-toplevel` in `fetchGitRoot` could be invoked through a function named `findTopLevelOfRepository`.

PR #555, opened 2021-12-29
ci: do not allow overwrite

drazisil-codecov:
do not replace published github release if tag exists


mitchell-codecov (approved):
We should additionally configure this on the GCP side of things.

PR #557, opened 2021-12-30
fix: use logging class for most verbose calls

Flamefire:
All operations that should depend on the set logging level must go through UploadLogger. The free functions are still required e.g. for things like `verbose(`Start of uploader: ${start}...`, realArgs.verbose)` running outside of `main`, so I kept them.

Consistently using `UploadLogger.verbose` is IMO more readable and makes it clear, that logging levels are used. If the free functions are to be preferred, the class can be removed.

To avoid mistakes with the free `verbose` remove the default Argument as that doesn't make sense anyway and is the source of current bugs. Those are fixed by this change as they all go through `UploadLogger` now.

@drazisil-codecov This is basically the all-round fix for https://github.com/codecov/uploader/pull/533#issuecomment-1002999737

The issue was likely introduced by some refactoring as evident in the test suite: `Logger Helper - Legacy log() tests'`

### DIFF
--- a/bin/codecov.ts
+++ b/bin/codecov.ts
@@ -15,7 +15,7 @@ const realArgs = argv.argv
 
 const start = Date.now()
 
-verbose(`Start of uploader: ${start}...`, Boolean(realArgs.verbose))
+verbose(`Start of uploader: ${start}...`, realArgs.verbose)
 main(realArgs)
   .then(() => {
     const end = Date.now()
--- a/src/ci_providers/provider_githubactions.ts
+++ b/src/ci_providers/provider_githubactions.ts
@@ -4,7 +4,7 @@
 import { IServiceParams, UploaderEnvs, UploaderInputs } from '../types'
 
 import { runExternalProgram } from "../helpers/util"
-import { info, verbose } from '../helpers/logger'
+import { info, UploadLogger } from '../helpers/logger'
 
 export function detect(envs: UploaderEnvs): boolean {
   return Boolean(envs.GITHUB_ACTIONS)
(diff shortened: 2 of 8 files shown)


Flamefire:
Decreased coverage not an issue, as many lines could be reduced. E.g. by proper Singleton usage (replace `init()` by `getInstance()`)

bin/codecov.ts
```diff
@@ -15,7 +15,7 @@ const realArgs = argv.argv
 
 const start = Date.now()
 
-verbose(`Start of uploader: ${start}...`, Boolean(realArgs.verbose))
+verbose(`Start of uploader: ${start}...`, realArgs.verbose)
```
drazisil-codecov, line 18: This was being coerced into a Boolean because `tsc` assumed it was a string and was throwing errors. If it now passes without it, 👍

drazisil-codecov (approved):
Agree this is much clearer, thanks!

Flamefire:
> @drazisil-codecov changed the title Fix logging system fix: use logging class for most verbose calls

Note that this fixes actual bugs where verbose messages were never logged due to the confusion caused by the different functions. Hence the more generic title

Flamefire:
> Please use conventional commits format in the future, otherwise, your changes will not populate to the changelog (and throw tons of errors if the repo's pre-commit hook ever sees them). Are you running npm install? The husky hook should be catching this. 🤔

What are "conventional commits format"? And no, not running npm install, don't even have the whole node toolchain on my PC. 

Why isn't that a CI check if the format is important? Would probably be worth adding.

PR #577, opened 2022-01-13
refactor: allow users to bypass blocklist with -f

drazisil-codecov:
In some cases, users would like to upload files that have names that would normally be blocked. If the file is passed with -f, we now emit a warning and allow it.


michael-codecov (approved):

src/helpers/files.ts
```diff
@@ -335,7 +335,15 @@ export function getBlocklist(): string[] {
   return [...manualBlocklist(), ...globBlocklist()].map(globstar)
 }
 
-export function cleanCoverageFilePaths(projectRoot: string, paths: string[], ignoreGlobs: string[]): string[] {
+export function filterFilesAgainstBlockList(paths: string[], ignoreGlobs: string[]): string[] {
+  const ignoredFiles = micromatch(paths, ignoreGlobs)
+
+  return paths.filter(path => {
+    return !ignoredFiles.includes(path)
+  })
```
mitchell-codecov, line 343:
```suggestion
  return micromatch.not(paths, ignoreGlobs, {
    contains: true,
  })
```

Disregard the line with `contains` if the contents of `ignoreGlobs` are already globstar'd.

ben-spiller:
Not sure if this is a regression but it seems in the same area as this fix: GH-590

PR #593, opened 2022-01-21
chore: vendor patched micromatch module

drazisil-codecov:
I'm not happy with this until I figure out why the tests passed before. 🤔


src/vendor/micromatch/index.js
```diff
@@ -0,0 +1,481 @@
...
+  let onResult = state => {
+    items.add(state.output)
+    if (options && options.onResult) {
+      options.onResult(state)
+    }
+  }
+
+  if (patterns.length === 0) {
+    return [...list]
+  }
```
drazisil-codecov, line 45: @mitchell-codecov I believe this was the only change.

mitchell-codecov (approved):

PR #602, opened 2022-01-27
Add gcov support

michael-codecov:


drazisil-codecov (approved):

src/helpers/gcov.ts
```diff
@@ -0,0 +1,23 @@
...
+import { isProgramInstalled, runExternalProgram } from "./util"
+
+export function generateGcovCoverageFiles(projectRoot: string, include: string[] = [], ignore: string[] = [], gcovArgs = '',) {
+    if (!isProgramInstalled('gcov')) {
+        throw new Error('gcov is not installed, cannot process files')
+    }
+
+    const globstar = (pattern: string) => `**/${pattern}`
+    const gcovInclude = ['*.gcno', ...include].map(globstar)
+    const gcovIgnore = ['bower_components', 'node_modules', 'vendor', ...ignore].map(globstar)
```
mitchell-codecov, line 11: Would it be worth using the results from one of the blocklist functions ([manualBlocklist], [globBlocklist])?

[manualBlocklist]: https://github.com/codecov/uploader/blob/701c7236bd1ea4ec258c1047654993ae881e46f9/src/helpers/files.ts#L30
[globBlocklist]: https://github.com/codecov/uploader/blob/701c7236bd1ea4ec258c1047654993ae881e46f9/src/helpers/files.ts#L46

src/helpers/gcov.ts
```diff
@@ -0,0 +1,23 @@
+import glob from 'fast-glob'
+import { isProgramInstalled, runExternalProgram } from "./util"
+
+export function generateGcovCoverageFiles(projectRoot: string, include: string[] = [], ignore: string[] = [], gcovArgs = '',) {
```
mitchell-codecov, line 4:
```suggestion
export async function generateGcovCoverageFiles(projectRoot: string, include: string[] = [], ignore: string[] = [], gcovArgs = '',) {
```

mitchell-codecov, line 4: The function should have its return type explicitly noted:
- improves API boundary by establishing a clear public contract
- provides self-documentation

src/helpers/gcov.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (!isProgramInstalled('gcov')) {
+        throw new Error('gcov is not installed, cannot process files')
+    }
+
+    const globstar = (pattern: string) => `**/${pattern}`
+    const gcovInclude = ['*.gcno', ...include].map(globstar)
+    const gcovIgnore = ['bower_components', 'node_modules', 'vendor', ...ignore].map(globstar)
+    const files = glob
+        .sync(gcovInclude, {
```
mitchell-codecov, line 13:
```suggestion
    const files = await glob(gcovInclude, {
```

Asynchrony should be used when possible.[^1]

Other instances of `glob` throughout the project should conform to this change accordingly.

[^1]: https://nodejs.org/en/knowledge/getting-started/control-flow/how-to-write-asynchronous-code/

src/helpers/gcov.ts
```diff
@@ -0,0 +1,23 @@
...
+            cwd: projectRoot,
+            ignore: gcovIgnore,
+            onlyFiles: true,
+        })
+    if (!files.length) {
+        throw new Error('No gcov files found')
+    }
+    const gcovLog = runExternalProgram('gcov', ['-pb', gcovArgs, files.join(' ')]);
```
mitchell-codecov, line 21: `gcovArgs` and `files` will each be passed in as one argument to gcov. Is this intended?

src/index.ts
```diff
@@ -153,6 +155,13 @@ export async function main(
   let requestedPaths: string[] = []
   
   // Look for files
+
+  if (args.gcov) {
+    UploadLogger.verbose('Running gcov...')
+
+    const gcovLogs = generateGcovCoverageFiles(projectRoot)
```
mitchell-codecov, line 162:
```suggestion
    const gcovLogs = await generateGcovCoverageFiles(projectRoot)
```

src/helpers/gcov.ts
michael-codecov: Not sure what you mean here, they are passed into `runExternalProgram` using the `optionalArguments` field.

src/helpers/gcov.ts
michael-codecov: I used the same 3 defaults as in bash, but I see all 3 are inside manualBlocklist so I will swap to that

src/helpers/gcov.ts
mitchell-codecov:
```suggestion
    const gcovLog = runExternalProgram('gcov', ['-pb', gcovArgs, ...files]);
```

There may be nuance with the way `gcovArgs` is passed. While it's a string, it may actually be treated as _one_ string in the context of it being an argument to gcov. For example:

```
const gcovArgs = "-B main -d --verbose";
// %expected-type: string

gcovArgs;
// => "-B main -d --verbose"
```
```
const gcovArgs = "-B main -d --verbose".split(" ");
// %expected-type: string[]

gcovArgs[0];
// => "-B"
```

src/index.ts
```diff
@@ -153,6 +155,17 @@ export async function main(
...
+
+  if (args.gcov) {
+    UploadLogger.verbose('Running gcov...')
+    try {
+      const gcovLogs = await generateGcovCoverageFiles(projectRoot)
+      UploadLogger.verbose(`${gcovLogs}`)
+    }
+    catch (error) {
+      throw new Error(`${error}`)
+    }
```
mitchell-codecov, line 167:
```suggestion
    const gcovLogs = await generateGcovCoverageFiles(projectRoot)
    UploadLogger.verbose(`${gcovLogs}`)
```

Allow the original error to fall through to the appropriate error handler.

michael-codecov: Hmm earlier I was getting an UnhandledPromiseRejection which is why I added this.  But with the test changes that seems to have disappeared

mitchell-codecov (approved):

PR #649, opened 2022-03-02
Fix detection of merge commits on Github Actions an AzP

Flamefire:
The double quotes get escaped by execFileSync and then are part of the output which makes the Regex fail to match.
Remove them to avoid this.

See https://github.com/boostorg/boost-ci/runs/5394406101?check_suite_focus=true:

```
['verbose'] Handling PR with parent hash(es) '"86f4b7cb20e29080b366ee1593c1135038bda4bf cd099c05fc6d88115e9bbdc0c60b23fb5a0fa952"' of current commit.
['info']     Commit with SHA 9e556cae25da2cb0438a10e81d3f78ddcc7e8510 of PR 125 is not a merge commit
```

Can be reproduced in a shell by running `git show --no-patch --format=\"%P\"`

@drazisil-codecov Finally found this!

### DIFF
--- a/src/ci_providers/provider_azurepipelines.ts
+++ b/src/ci_providers/provider_azurepipelines.ts
@@ -60,7 +60,7 @@ function _getSHA(inputs: UploaderInputs): string {
   if (_getPR(inputs)) {
     const mergeCommitRegex = /^[a-z0-9]{40} [a-z0-9]{40}$/
     const mergeCommitMessage = childProcess
-      .execFileSync('git', ['show', '--no-patch', '--format="%P"'])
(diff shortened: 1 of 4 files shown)


drazisil-codecov (approved):

drazisil-codecov (changes requested):
Tests appear to be failing after pulling in main, @Flamefire 
Can you take a look?

Flamefire:
> Tests appear to be failing after pulling in main, @Flamefire Can you take a look?

That merge actually exposed some missing mocks. While fixing this I found another bug in an affected test and fixed that too as well as the commit messages (hence the force-push)

drazisil-codecov (approved):

PR #714, opened 2022-04-13
feat: add coveragepy xml support

thomasrockhu-codecov:
Adds `coveragepy` support, it automagic runs `coverage xml`

### DIFF
--- a/src/helpers/coveragepy.ts
+++ b/src/helpers/coveragepy.ts
@@ -1,0 +1,9 @@
+import { isProgramInstalled, runExternalProgram } from "./util"
+
+export async function generateCoveragePyFile(): Promise<string> {
+    if (!isProgramInstalled('coverage')) {
+        throw new Error('coveragepy is not installed, cannot convert file format')
+    }
+
+    return runExternalProgram('coverage', ['xml']);
+}
--- a/src/helpers/coveragepy.ts
+++ b/src/helpers/coveragepy.ts
@@ -1,11 +1,11 @@
 import { isProgramInstalled, runExternalProgram } from "./util"
-import { UploadLogger } from './logger'
+import { info } from './logger'
 
 export async function generateCoveragePyFile(): Promise<string> {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -165,7 +166,11 @@ export async function main(
     const gcovLogs = await generateGcovCoverageFiles(projectRoot, gcovInclude, gcovIgnore, gcovArgs)
     UploadLogger.verbose(`${gcovLogs}`)
   }
-  
+
+  UploadLogger.verbose('Running coverage xml...')
+  const coveragePyLogs = await generateCoveragePyFile()
+  UploadLogger.verbose(`${coveragePyLogs}`)
```
mitchell-codecov, line 172: Should this be behind a feature flag of some kind?

thomasrockhu-codecov: No, I don't think so, I think we should be doing this everytime

src/helpers/coveragepy.ts
```diff
@@ -0,0 +1,11 @@
+import { isProgramInstalled, runExternalProgram } from "./util"
+import { UploadLogger } from './logger'
+
+export async function generateCoveragePyFile(): Promise<string> {
+    if (!isProgramInstalled('coverage')) {
+        UploadLogger.verbose(`coveragepy is not installed, skipping coverage xml conversion`)
+        return '';
```
mitchell-codecov, line 7:
```suggestion
        throw new Error('coveragepy is not installed, skipping coverage xml conversion')
```

https://github.com/codecov/uploader/blob/469569cd1d39e906225010318051294eb2923f2f/src/helpers/xcode.ts#L12

mitchell-codecov: A try-catch may be of use back in index.ts where the function is called.

mitchell-codecov (changes requested):
https://github.com/codecov/uploader/pull/714#discussion_r853161047

mitchell-codecov (approved):

PR #715, opened 2022-04-14
Feat xcode

thomasrockhu-codecov:
Note that `xcode.ts` is basically a re-write from our [poc](https://github.com/codecov/xcode-poc/blob/master/generate-codecov-json.js)

### DIFF
--- a/src/ci_providers/provider_azurepipelines.ts
+++ b/src/ci_providers/provider_azurepipelines.ts
@@ -66,7 +66,7 @@ function _getSHA(inputs: UploaderInputs): string {
     if (mergeCommitRegex.exec(mergeCommitMessage)) {
       const mergeCommit = mergeCommitMessage.split(' ')[1]
       info(`    Fixing merge commit SHA ${commit} -> ${mergeCommit}`)
-      commit = mergeCommit
+      commit = mergeCommit || ''
     }
   }
 
--- a/src/ci_providers/provider_circleci.ts
+++ b/src/ci_providers/provider_circleci.ts
@@ -43,7 +43,7 @@ function _getSlug(inputs: UploaderInputs): string {
   }
 
   if (isSetAndNotEmpty(envs.CIRCLE_REPOSITORY_URL)) {
-    return `${envs.CIRCLE_REPOSITORY_URL?.split(':')[1].split('.git')[0]}`
+    return `${envs.CIRCLE_REPOSITORY_URL?.split(':')[1]?.split('.git')[0]}`
   }
   return slug
 }
--- a/src/helpers/cli.ts
+++ b/src/helpers/cli.ts
@@ -186,7 +186,7 @@ const args: ICLIArgument[] = [
     description: 'Extra arguments to pass to gcov',
   },
   {
-    alias: 'x',
+    alias: 'xc',
     name: 'xcode',
     type: 'boolean',
     default: false,
--- a/src/helpers/cli.ts
+++ b/src/helpers/cli.ts
@@ -185,6 +185,19 @@ const args: ICLIArgument[] = [
     type: 'string',
     description: 'Extra arguments to pass to gcov',
   },
+  {
+    alias: 'x',
+    name: 'xcode',
+    type: 'boolean',
+    default: false,
+    description: 'Run with xcode support',
+  },
+  {
+    alias: 'xp',
+    name: 'xcodeArchivePath',
+    type: 'string',
+    description: 'Specify the xcode archive path. Likely specified as the -resultBundlePath and should end in .xcresult',
+  },
 ]
 
 export interface IYargsObject {
--- a/src/helpers/git.ts
+++ b/src/helpers/git.ts
@@ -9,14 +9,14 @@ export function parseSlug(slug: string): string {
 
   if (slug.match('http')) {
     // Type is http(s)
-    const phaseOne = slug.split('//')[1].replace('.git', '')
-    const phaseTwo = phaseOne.split('/')
+    const phaseOne = slug.split('//')[1]?.replace('.git', '') || ''
+    const phaseTwo = phaseOne?.split('/') || ''
     const cleanSlug = `${phaseTwo[1]}/${phaseTwo[2]}`
     return cleanSlug
   } else if (slug.match('@')) {
     // Type is git
-    const cleanSlug = slug.split(':')[1].replace('.git', '')
-    return cleanSlug
+    const cleanSlug = slug.split(':')[1]?.replace('.git', '')
+    return cleanSlug || ''
   }
   throw new Error(`Unable to parse slug URL: ${slug}`)
 }
--- a/src/helpers/xcode.ts
+++ b/src/helpers/xcode.ts
@@ -1,4 +1,4 @@
-import fs from 'fs'
+import fs from 'fs/promises'
 
 import { info, UploadLogger } from '../helpers/logger'
 import {
@@ -29,7 +29,7 @@ export async function generateXcodeCoverageFiles(archivePath: string): Promise<s
   }
   const filename = `./coverage-report-${pathFilename}.json`
   UploadLogger.verbose(`Writing coverage to ${filename}`)
-  fs.writeFileSync(filename, JSON.stringify(report))
+  await fs.writeFile(filename, JSON.stringify(report))
   return filename
 }
 
@@ -42,13 +42,16 @@ function getCoverageInfo(archivePath: string, filePath: string): string {
   return runExternalProgram('xcrun', ['xccov', 'view', '--archive', archivePath, '--file', filePath])
 }
 
-function convertCoverage (coverageInfo: string): XcodeCoverageFileReport {
+function convertCoverage(coverageInfo: string): XcodeCoverageFileReport {
   const coverageInfoArr = coverageInfo.split('\n')
   const obj: XcodeCoverageFileReport = {}
   coverageInfoArr.forEach(line => {
     const [lineNum, lineInfo] = line.split(':')
     if (lineNum && Number.isInteger(Number(lineNum))) {
-      const lineHits = lineInfo.trimStart().split(' ')[0].trim()
+      const lineHits = lineInfo?.trimStart().split(' ')[0]?.trim()
+      if (typeof lineHits !== 'string') {
+        return
+      }
       if (lineHits === '*') {
         obj[String(lineNum.trim())] = null
       } else {
--- a/src/helpers/xcode.ts
+++ b/src/helpers/xcode.ts
@@ -39,7 +39,7 @@ function getFileList(archivePath: string): string[] {
   return fileList.split('\n').filter(i => i !== '')
 }
 
-function getCoverageInfo (archivePath: string, filePath: string): string {
+function getCoverageInfo(archivePath: string, filePath: string): string {
   return runExternalProgram('xcrun', ['xccov', 'view', '--archive', archivePath, '--file', filePath])
 }
 
--- a/src/helpers/xcode.ts
+++ b/src/helpers/xcode.ts
@@ -15,16 +15,15 @@ export async function generateXcodeCoverageFiles(archivePath: string): Promise<s
 
   const coverage: XcodeCoverageReport = {}
   const report = { coverage: coverage }
-  const path = archivePath[0]
 
-  getFileList(path).forEach(repoFilePath => {
+  getFileList(archivePath).forEach(repoFilePath => {
     UploadLogger.verbose(`Converting ${repoFilePath}...`)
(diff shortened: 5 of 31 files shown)


src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+  fs.writeFileSync(filename, JSON.stringify(report))
+  return filename
+}
+
+function getFileList(archivePath: string): string[] {
+  const fileList = runExternalProgram('xcrun', ['xccov', 'view', '--file-list', '--archive', archivePath]);
+  return fileList.split('\n').filter(i => i !== '')
+}
+
+function getCoverageInfo (archivePath: string, filePath: string): string {
```
mitchell-codecov, line 42:
```suggestion
function getCoverageInfo(archivePath: string, filePath: string): string {
```

src/types.ts
```diff
@@ -81,4 +83,7 @@ export interface PostResults {
 export interface PutResults {
   status: string
   resultURL: URL
-}
\ No newline at end of file
+}
+
+export type XcodeCoverageFileReport = NodeJS.Dict<string>
+export type XcodeCoverageReport = NodeJS.Dict<XcodeCoverageFileReport>
```
mitchell-codecov, line 89:
```suggestion
export type XcodeCoverageFileReport = Record<string, string>
export type XcodeCoverageReport = Record<string, XcodeCoverageFileReport>
```

mitchell-codecov: `UploaderEnvs` should also be updated according to the suggested changes in a separate PR.

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export async function generateXcodeCoverageFiles(archivePath: string[]): Promise<string> {
+  if (!isProgramInstalled('xcrun')) {
+      throw new Error('xcrun is not installed, cannot process files')
+  }
+  info('Running xcode coversion...')
+
+  const coverage: XcodeCoverageReport = {}
+  const report = { coverage: coverage }
+  const path = archivePath[0]
```
mitchell-codecov, line 18: Only one path from `archivePath` is ever used. Is this intentional?

thomasrockhu-codecov: It comes from https://github.com/codecov/uploader/pull/715/files/89e3f7d4c126d68852a2292e70a99694dbdf913f#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R174, and I was struggling to wrangle the types here.

Not sure if there's a better way to do this

src/types.ts
thomasrockhu-codecov: @mitchell-codecov sometimes it resolves to `undefined` I think

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+  return runExternalProgram('xcrun', ['xccov', 'view', '--archive', archivePath, '--file', filePath])
+}
+
+function convertCoverage (coverageInfo: string): XcodeCoverageFileReport {
+  const coverageInfoArr = coverageInfo.split('\n')
+  const obj: XcodeCoverageFileReport = {}
+  coverageInfoArr.forEach(line => {
+    const [lineNum, lineInfo] = line.split(':')
+    if (lineNum && Number.isInteger(Number(lineNum))) {
+      const lineHits = lineInfo.trimStart().split(' ')[0].trim()
```
mitchell-codecov, line 52:
```suggestion
      const lineHits = lineInfo?.trimStart().split(' ')[0]?.trim()
```

With the above suggestion, `lineHits` would be type `string | undefined`. An additional conditional statement somewhere below would preserve the indexing logic. For example:

```ts
if (typeof lineHits !== "string") {
  return
}
```

mitchell-codecov: The enablement of `noUncheckedIndexedAccess` in tsconfig.json would have improved IDE typing here and made the above observations more apparent.

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (lineHits === '*') {
+        obj[String(lineNum.trim())] = undefined
```
mitchell-codecov, line 54:
```suggestion
        obj[String(lineNum.trim())] = null
```

`null` shows clear intent.

mitchell-codecov: The type definition of `XcodeCoverageFileReport` should be modified to accommodate for line numbers without coverage information.

mitchell-codecov (changes requested):

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+function getCoverageInfo (archivePath: string, filePath: string): string {
+  return runExternalProgram('xcrun', ['xccov', 'view', '--archive', archivePath, '--file', filePath])
+}
+
+function convertCoverage (coverageInfo: string): XcodeCoverageFileReport {
+  const coverageInfoArr = coverageInfo.split('\n')
+  const obj: XcodeCoverageFileReport = {}
+  coverageInfoArr.forEach(line => {
+    const [lineNum, lineInfo] = line.split(':')
+    if (lineNum && Number.isInteger(Number(lineNum))) {
```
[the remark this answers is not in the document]
thomasrockhu-codecov: That's fine for now, I believe xcode spits back 5 to us

src/helpers/xcode.ts
mitchell-codecov: The use of the String constructor here is extraneous since String.prototype.trim() already returns a string.[^1]

[^1]: [String.prototype.trim() - JavaScript | MDN]

[String.prototype.trim() - JavaScript | MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+    report.coverage[repoFilePath] = coverageJson
+  })
+
+  let pathFilename = path.split('/').pop()
+  if (pathFilename) {
+    pathFilename = pathFilename.split('.xcresult')[0]
+  }
+  const filename = `./coverage-report-${pathFilename}.json`
+  UploadLogger.verbose(`Writing coverage to ${filename}`)
+  fs.writeFileSync(filename, JSON.stringify(report))
```
mitchell-codecov, line 33:
```suggestion
  await fs.writeFile(filename, JSON.stringify(report))
```

Be sure to modify the import statement to load the `fs/promises` module instead of the `fs`.

src/helpers/xcode.ts
```diff
@@ -0,0 +1,61 @@
...
+    UploadLogger.verbose(`Converting ${repoFilePath}...`)
+    const coverageInfo = getCoverageInfo(path, repoFilePath)
+    const coverageJson = convertCoverage(coverageInfo)
+    report.coverage[repoFilePath] = coverageJson
+  })
+
+  let pathFilename = path.split('/').pop()
+  if (pathFilename) {
+    pathFilename = pathFilename.split('.xcresult')[0]
+  }
```
mitchell-codecov, line 30: Should there be an `else` to this? `pathFilename` can be `undefined` after line 27.

src/helpers/xcode.ts
```diff
@@ -0,0 +1,60 @@
...
+function getFileList(archivePath: string): string[] {
+  const fileList = runExternalProgram('xcrun', ['xccov', 'view', '--file-list', '--archive', archivePath]);
+  return fileList.split('\n').filter(i => i !== '')
+}
+
+function getCoverageInfo(archivePath: string, filePath: string): string {
+  return runExternalProgram('xcrun', ['xccov', 'view', '--archive', archivePath, '--file', filePath])
+}
+
+function convertCoverage (coverageInfo: string): XcodeCoverageFileReport {
```
mitchell-codecov, line 45:
```suggestion
function convertCoverage(coverageInfo: string): XcodeCoverageFileReport {
```

src/helpers/cli.ts
```diff
@@ -185,6 +185,19 @@ const args: ICLIArgument[] = [
     type: 'string',
     description: 'Extra arguments to pass to gcov',
   },
+  {
+    alias: 'xc',
```
mitchell-codecov, line 189: Keep in mind that multicharacter short flags do not work with yargs.

thomasrockhu-codecov: 💯

src/helpers/xcode.ts
thomasrockhu-codecov: Out of curiosity, why is this necessary?

src/helpers/xcode.ts
thomasrockhu-codecov: no, this is going to be `''` at worst.

src/helpers/xcode.ts
mitchell-codecov: There are very strong benefits and very few drawbacks to using an asynchronous version of a function over its synchronous counterpart.

`writeFile` is asynchronous and, as a result, does not block the execution of the program. Node.js can schedule and run other parts of the program _while_ the file is being written to. For the uploader in particular, this may mean very little in its current state; however, if other areas of it were written to take serious advantage of asynchronous functionality, further asynchrony would be synergetic with the overall performance of the application. Function calls could weave together and through each other.

src/ci_providers/provider_azurepipelines.ts
```diff
@@ -66,7 +66,7 @@ function _getSHA(inputs: UploaderInputs): string {
     if (mergeCommitRegex.exec(mergeCommitMessage)) {
       const mergeCommit = mergeCommitMessage.split(' ')[1]
       info(`    Fixing merge commit SHA ${commit} -> ${mergeCommit}`)
-      commit = mergeCommit
+      commit = mergeCommit || ''
```
mitchell-codecov, line 69:
```suggestion
      commit = mergeCommit ?? ''
```

The use of `??` circumvents some scenarios where `||` would consider the first operand to be false.[^1]

[^1]: [Nullish coalescing operator (??) - JavaScript | MDN]

[Nullish coalescing operator (??) - JavaScript | MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

src/ci_providers/provider_circleci.ts
```diff
@@ -43,7 +43,7 @@ function _getSlug(inputs: UploaderInputs): string {
   }
 
   if (isSetAndNotEmpty(envs.CIRCLE_REPOSITORY_URL)) {
-    return `${envs.CIRCLE_REPOSITORY_URL?.split(':')[1].split('.git')[0]}`
+    return `${envs.CIRCLE_REPOSITORY_URL?.split(':')[1]?.split('.git')[0]}`
```
mitchell-codecov, line 46: 🙌

src/helpers/git.ts
```diff
@@ -9,14 +9,14 @@ export function parseSlug(slug: string): string {
 
   if (slug.match('http')) {
     // Type is http(s)
-    const phaseOne = slug.split('//')[1].replace('.git', '')
-    const phaseTwo = phaseOne.split('/')
+    const phaseOne = slug.split('//')[1]?.replace('.git', '') || ''
+    const phaseTwo = phaseOne?.split('/') || ''
```
mitchell-codecov, line 13:
```suggestion
    const phaseOne = slug.split('//')[1]?.replace('.git', '') ?? ''
    const phaseTwo = phaseOne?.split('/') ?? ''
```

src/helpers/git.ts
```diff
@@ -9,14 +9,14 @@ export function parseSlug(slug: string): string {
...
     const cleanSlug = `${phaseTwo[1]}/${phaseTwo[2]}`
     return cleanSlug
   } else if (slug.match('@')) {
     // Type is git
-    const cleanSlug = slug.split(':')[1].replace('.git', '')
-    return cleanSlug
+    const cleanSlug = slug.split(':')[1]?.replace('.git', '')
+    return cleanSlug || ''
```
mitchell-codecov, line 19:
```suggestion
    return cleanSlug ?? ''
```

src/index.ts
```diff
@@ -167,6 +168,16 @@ export async function main(
...
+  if (args.xcode) {
+    if (!args.xcodeArchivePath) {
+      throw new Error('Please specify xcodeArchivePath to run the Codecov uploader with xcode support')
+    } else {
+      const xcodeArchivePath: string = args.xcodeArchivePath
+      const xcodeLogs = await generateXcodeCoverageFiles(xcodeArchivePath)
+      UploadLogger.verbose(`${xcodeLogs}`)
+    }
+  }
+
```
mitchell-codecov, line 180: 🙌

mitchell-codecov (approved):

PR #716, opened 2022-04-14
fix: increase spawnsync buffer

thomasrockhu-codecov:
Fixes the issue [here](https://github.com/RemedyIT/taox11/runs/5996166946?check_suite_focus=true) where running `gcov` exceeds the buffer size. Also removes unused argument from `token.ts`

### DIFF
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -1,6 +1,6 @@
 import childprocess from 'child_process'
 
-export const SPAWNPROCESSBUFFERSIZE = 10485760 // 10 MiB
+export const SPAWNPROCESSBUFFERSIZE: number = 10485760 // 10 MiB
 
 export function isProgramInstalled(programName: string): boolean {
   return !childprocess.spawnSync(programName).error
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -1,6 +1,6 @@
 import childprocess from 'child_process'
 
-export const SPAWNPROCESSBUFFERSIZE: number = 10485760 // 10 MiB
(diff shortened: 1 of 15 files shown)


src/helpers/util.ts
```diff
@@ -8,7 +10,14 @@ export function runExternalProgram(
...
 ): string {
-  const result = childprocess.spawnSync(programName, optionalArguments)
+  const result = childprocess.spawnSync(
+    programName,
+    optionalArguments,
+    { maxBuffer: SPAWNPROCESSBUFFERSIZE },
+  )
+  if (result === undefined) {
+    throw new Error(`${programName}, ${typeof optionalArguments}, ${Array.isArray(optionalArguments)}, ${optionalArguments}`)
+  }
```
mitchell-codecov, line 20: What causes `result` to be `undefined`?

thomasrockhu-codecov: @mitchell-codecov my latest change didn't push this file 😬. It was supposed to be removed

src/helpers/util.ts
```diff
@@ -1,5 +1,7 @@
 import childprocess from 'child_process'
 
+export const SPAWNPROCESSBUFFERSIZE = 10485760 // 10 MiB
```
mitchell-codecov:
```suggestion
export const SPAWNPROCESSBUFFERSIZE = 1_048_576 * 10 // 10 MiB
```

(Readability) Underscores can improve readability in place of would-be commas.

Feel free to resolve this suggestion if you feel it does not improve readability.

thomasrockhu-codecov: @mitchell-codecov I think it's fine either way (especially because it's commented with 10MiB)

Issue #704, opened 2022-04-11
Codecov uploader OOMs while searching for file network on large repositories


gabrielrussoc:
**Describe the bug**
Running the uploader on large repositories causes it to OOM while discovering the file network.

```
[2022-04-11T12:39:42.234Z] ['info'] 
     _____          _
    / ____|        | |
   | |     ___   __| | ___  ___ _____   __
   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
   | |___| (_) | (_| |  __/ (_| (_) \ V /
    \_____\___/ \__,_|\___|\___\___/ \_/

  Codecov report uploader 0.1.20
[2022-04-11T12:39:42.240Z] ['info'] => Project root located at: /ephemeral/home/ubuntu/worker_workspace
[2022-04-11T12:39:42.242Z] ['info'] ->  Token found by environment variables
[2022-04-11T12:39:42.242Z] ['info'] ->  Token set by environment variables

<--- Last few GCs --->

[22463:0x7fe065602300]   168398 ms: Mark-sweep 4052.5 (4135.6) -> 4042.5 (4141.9) MB, 2926.0 / 0.0 ms  (average mu = 0.584, current mu = 0.255) allocation failure scavenge might not succeed
[22463:0x7fe065602300]   172823 ms: Mark-sweep 4060.3 (4143.6) -> 4049.4 (4148.6) MB, 4104.4 / 0.0 ms  (average mu = 0.367, current mu = 0.072) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

I'm forced to used `-X network` to unblock.

**To Reproduce**
1. On a very large repository (I couldn't find any public one)
2. Run the uploader binary without any args or environment (I'm on Ubuntu 18):
`./codecov`

**Expected behavior**
Uploader should not OOM

**Screenshots**
n/a

**Additional context**
I did some digging and the problem seems to be exactly here: https://github.com/codecov/uploader/blob/2ad72d4c48aed30338c00e848d856c481beee7ba/src/helpers/files.ts#L238-L263

[...]

PR #729, opened 2022-04-21
fix: increase maxBuffer of spawnSync

gabrielrussoc:
nodeJS limits the size of the stdout of child processes to 1MB: https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options

The output of `git ls-files` exceeds 1MB in my case, falling back to using `glob`. However the glob call [hardcodes](https://github.com/codecov/uploader/blob/2ad72d4c48aed30338c00e848d856c481beee7ba/src/helpers/files.ts#L249) a list of files that doesn't respect .gitignore and causes weird behaviour.

I'm setting 100MB as an arbitrary large value. There is some discussion about the default value of this argument here: https://github.com/nodejs/node/issues/9829

Motivated by https://github.com/codecov/uploader/issues/704

## Tests

I tried to find tests for this behaviour but it looks like there are none. I manually ran the uploader on a large repository with more than 1MB output:

```
$ git ls-files > ls_files.txt
$ du -sh ls_files.txt
6.8M	ls_files.txt
```

**Before**:
```
./uploader ...
...
<--- Last few GCs --->

[22463:0x7fe065602300]   168398 ms: Mark-sweep 4052.5 (4135.6) -> 4042.5 (4141.9) MB, 2926.0 / 0.0 ms  (average mu = 0.584, current mu = 0.255) allocation failure scavenge might not succeed
[22463:0x7fe065602300]   172823 ms: Mark-sweep 4060.3 (4143.6) -> 4049.4 (4148.6) MB, 4104.4 / 0.0 ms  (average mu = 0.367, current mu = 0.072) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

**After**:
```
./uploader ...
...
[2022-04-21T20:58:15.604Z] ['info'] Uploading...
[2022-04-21T20:58:17.644Z] ['info'] {"status":"success","resultURL":""}
```


gabrielrussoc:
I can confirm the issue still reproes on 0.2.0, looks like #716 missed this call to `spawnSync`

gabrielrussoc:
Here's a public repro:

On the linux repository: https://github.com/torvalds/linux (has more than 1MB files)

```
$ ln -s . INFINITE_LOOP_TO_BREAK_CODECOV

$ ./codecov-linux
```

The repo is too large, `git ls-files` fails, it falls back to glob, glob doesn't ignore the circular symlink, memory explodes

gabrielrussoc:
I believe this is used to discover the file network and not to look for reports

drazisil-codecov (approved):

PR #756, opened 2022-05-06
Jenkins: branch/commit handling for multibranch pipelines

M1ke:
When using multibranch pipelines, Jenkins will create a new branch for pull requests named "PR-X" where X is the ID. This means that submitting the branch name to Codecov will not represent real branches from the project.

It is also the case that in multibranch projects, Jenkins may merge `master` into the working branch. In the event this creates a merge commit the `GIT_COMMIT` envvar will point to this rather than the commit being tested (and thus the given commit hash will be irrelevant when posting to external services). This PR does not fix that but adds a note in the code to assist future people encountering this issue.


drazisil-codecov (approved):
Thank you for the PR 🙌

PR #762, opened 2022-05-12
Disable `standards-coverage-comparison`

mitchell-codecov:
Autotest has been failing for quite some time. While the failed CI runs do not have material impact on the project, the status checks they send to GitHub show up next to each commit. The failed indicator is misleading and a cause of information saturation.


drazisil-codecov (approved):
🙏

Issue #786, opened 2022-06-01
Specify which `gcov` executable to run


thomasrockhu-codecov:
**Is your feature request related to a problem? Please describe.**
The bash uploader allows for a specific `gcov` executable to be run `-x`

**Describe the solution you'd like**
Copy that functionality

**Describe alternatives you've considered**
/shrug

**Additional context**

mitchell-codecov:
The user should be able to specify which version of the executable is used by setting it earlier in their `PATH` environment variable. Is that the kind of functionality you're looking for or are you interested in something more akin npm's [`git`] configuration value?

[`git`]: https://docs.npmjs.com/cli/v8/using-npm/config#git

thomasrockhu-codecov:
@mitchell-codecov we hardcode `gcov`

https://github.com/codecov/uploader/blob/main/src/helpers/gcov.ts#L7

PR #794, opened 2022-06-04
feat: add gcov executable

thomasrockhu-codecov:
fixes https://github.com/codecov/uploader/issues/786


src/helpers/gcov.ts
```diff
@@ -15,5 +15,8 @@ export async function generateGcovCoverageFiles(projectRoot: string, include: st
     if (!files.length) {
         throw new Error('No gcov files found')
     }
-    return runExternalProgram('gcov', ['-pb', ...gcovArgs, ...files]);
+    if (gcovExecutable == 'gcov') {
+        gcovArgs.unshift('-pb')
```
mitchell-codecov, line 19: Where does `-pb` come from?

src/helpers/gcov.ts
```diff
@@ -15,5 +15,8 @@ export async function generateGcovCoverageFiles(projectRoot: string, include: st
     if (!files.length) {
         throw new Error('No gcov files found')
     }
-    return runExternalProgram('gcov', ['-pb', ...gcovArgs, ...files]);
+    if (gcovExecutable == 'gcov') {
```
mitchell-codecov, line 18:
```suggestion
    if (gcovExecutable === 'gcov') {
```

src/helpers/gcov.ts
mitchell-codecov: Thanks, Tom. I thought `-pb` was being _removed_ from the array rather than added to it.

I clearly need to brush up on my JavaScript!

mitchell-codecov (approved):

lukaszstolarczuk:
Hi, is there a chance this PR will be merged? :-)

lukaszstolarczuk:
> From https://en.wikipedia.org/wiki/Gcov#Coverage_summaries
> 
> > Lcov is a [graphical front-end](https://en.wikipedia.org/wiki/Graphical_user_interface) for gcov. It collects gcov data for multiple source files and creates [HTML](https://en.wikipedia.org/wiki/HTML) pages containing the source code annotated with coverage information.
> 
> Why would a customer use `lcov`? In addition to creating HTML files, which Codecov does not support, is it able to be executed without `gcov` being present? What situations would this be the case?

In our case it wasn't `lcov`, it was `-x "llvm-cov gcov"` with `clang` compilation. As far as I understand, they might produce different coverage reports from `gcov` + `gcc`.

lukaszstolarczuk:
> > In our case it wasn't `lcov`, it was `-x "llvm-cov gcov"` with `clang` compilation. As far as I understand, they might produce different coverage reports from `gcov` + `gcc`.
> 
> @lukaszstolarczuk ,
> 
> In that case, since you are calling `gcov` manually, wouldn't it work for you to not call the Codecov `gcov` and run the uploader normally? It should pick up your generated reports. Apologies if I do not quite understand your use case.

Oh no, we never call `gcov` manually. With bash uploader we used to call you script like this:
```bash
if [[ ${clang_used} -gt 0 ]]; then
	gcovexe="llvm-cov gcov"
else
	gcovexe="gcov"
fi
/opt/scripts/codecov -c -F ${1} -Z -x "${gcovexe}" "gcovout"
```

right now, with the new uploader, we just ignore that parameter, because it's missing 😉 
```bash
/opt/scripts/codecov --flags ${1} --nonZero --gcov --rootDir . --clean
```

We never confirmed what the reason was, but since the day we started using the new uploader, our coverage dropped by ~6% (the only difference we noticed for sure, was missing clang builds coverage reports).

mitchell-codecov (approved):

PR #805, opened 2022-06-15
Fix typo: mannually -> manually

dmivankov:


drazisil-codecov (approved):
🤦‍♀️ Good catch, @dmivankov, thank you!

PR #807, opened 2022-06-16
(DO NOT MERGE) fix: require non-empty patterns

### LINKED ISSUE #677: [CODE-1717] Version 0.1.20 treats `-f ''` as `-f '**/` and attempts to include all files
**Describe the bug**
Version 0.1.20 treats `-f ''` as `-f '**/` and attempts to include all files

**To Reproduce**
1. Pass `-f ''` to the uploader
2. Observe file list of coverage paths to clean

**Expected behavior**

Return a warning that `-f` was passed an empty string and delete the arg, behaving as if it was not passed.

**Screenshots**

**Additional context**
This is essentially painful on large repositories.

mitchell-codecov:
Fixes #677

### DIFF
--- a/src/helpers/files.ts
+++ b/src/helpers/files.ts
@@ -1,4 +1,4 @@
-import { default as assert } from 'assert/strict'
+import * as assert from 'assert'
 import { spawnSync } from 'child_process'
 import glob from 'fast-glob'
 import fs from 'fs'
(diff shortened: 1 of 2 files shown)


src/helpers/files.ts
```diff
@@ -1,3 +1,4 @@
+import { default as assert } from 'assert/strict'
```
mitchell-codecov, line 1:
```suggestion
import * as assert from 'assert'
```

drazisil-codecov (changes requested):
Please ensure this will not throw and will instead return an error and honor the exit code flag of -`Z`

PR #831, opened 2022-07-25
fix: throw error when flags do not match pattern

### LINKED ISSUE #829: Flag silently ignored when contains invalid characters
**Describe the bug**

>Flags Must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters

If my CI suite accidentally contains a slash or some other character, the flag is silently dropped.

**To Reproduce**
1. Use uploader v0.2.4
2. send a flag as `-F ci/js-unit-tests`
3. The upload succeeds but the flag is blank

**Expected behavior**

I would expect it to bail out with "Invalid flag: flags Must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters"

**Additional context**

If you think this is too much of a BC break, I could accept this being hidden behind a flag like `CODECOV_STRICT_FLAGS=1` or something.

drazisil-codecov:
Prior behavior was to strip out invalid flag names. The validation function will now throw.

Fixes #829


src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
...
   // eslint-disable-next-line no-useless-escape
   const mask = /^[\w\.\-]{1,45}$/
-  return mask.test(flags)
+  if (flag.length !== 0 && mask.test(flag) !== true) {
+    throw new Error(
+      `Flags must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters. Received ${flag}`,
+    )
+  }
+  return true
 }
```
mitchell-codecov, line 26:
```suggestion
export function isValidFlag(flag: string): boolean {
  return flag.length > 0 && mask.test(flag)
}

export function validateFlags(flags: string[]): void {
  const invalidFlags = flags.filter(isValidFlag)
  
  if (invalidFlags.length > 0) {
    throw new Error(
      `Flags must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters. Received ${flags.join(',')}`,
    )
  }
}
```

src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
   return validator.isURL(url, { require_protocol: true })
 }
 
-export function validateFlags(flags: string): boolean {
+export function validateFlags(flag: string): boolean {
```
mitchell-codecov, line 17:
```suggestion
export function validateFlags(flags: string[]): void {
```

src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
...
   // eslint-disable-next-line no-useless-escape
   const mask = /^[\w\.\-]{1,45}$/
```
mitchell-codecov, line 19:
```suggestion
```

src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
...
-  return mask.test(flags)
+  if (flag.length !== 0 && mask.test(flag) !== true) {
```
mitchell-codecov, line 20:
```suggestion
  const invalidFlags = flags.filter(isValidFlag)

  if (invalidFlags.length > 0) {
```

src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
...
+    throw new Error(
+      `Flags must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters. Received ${flag}`,
```
mitchell-codecov, line 22:
```suggestion
      `Flags must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters. Received ${flags}`,
```

src/helpers/validate.ts
```diff
@@ -14,10 +14,15 @@ export function validateURL(url: string): boolean {
...
+    )
+  }
+  return true
```
mitchell-codecov, line 25:
```suggestion
```

mitchell-codecov (changes requested):

src/helpers/validate.ts
drazisil-codecov: This is being passed a single flag as part of a filter command, I believe, @mitchell-codecov

src/helpers/validate.ts
drazisil-codecov: @mitchell-codecov Why are you deleting the mask?

mitchell-codecov (approved):

PR #848, opened 2022-08-15
Skip empty paths when more than one is passed

mitchell-codecov:


src/index.ts
```diff
@@ -193,12 +193,16 @@ export async function main(
   }
 
   let coverageFilePaths: string[] = []
-  if (args.file) {
+  if (args.file !== undefined) {
```
mitchell-codecov, line 196: This condition previously skipped empty paths if:
- only one occurrence of the `-f` flag was used; and
- the value passed to it was an empty string.

src/index.ts
```diff
@@ -193,12 +193,16 @@ export async function main(
...
     if (typeof args.file === 'string') {
       requestedPaths = [args.file]
     } else {
-      requestedPaths = args.file
+      requestedPaths = args.file // Already an array
     }
+
+    requestedPaths = requestedPaths.filter((path) => {
+      return Boolean(path) ?? info('Warning: Skipping an empty path passed to `-f`')
```
mitchell-codecov, line 204:
```suggestion
      return Boolean(path) || info('Warning: Skipping an empty path passed to `-f`')
```

The `??` operator does not short circuit if the first operand is falsy.

drazisil-codecov (approved):

PR #857, opened 2022-08-22
Don't call `checkSlug` if a token is passed

mitchell-codecov:
It's not necessary to validate `slug` if a token is passed in; the server can determine it from the upload token.


src/index.ts
```diff
@@ -342,9 +342,17 @@ export async function main(
     UploadLogger.verbose(`${parameter}`)
   }
 
-  const validSlug =  checkSlug(buildParams.slug)
-  if (!validSlug) {
-    throw new Error(`Slug must follow the format of "<owner>/<repo>" or be blank. We detected "${buildParams.slug}"`)
+  if (!hasToken) {
+    if (!buildParams.slug) {
+      throw new Error(`Slug must be set if a token is not passed`)
```
thomasrockhu, line 347: Maybe add a line on how to set the slug (or if it's an env variable)

mitchell-codecov, line 347:
```suggestion
      throw new Error(
        'Slug must be set if a token is not passed. Consider passing a slug via `-r`',
      )
```

mitchell-codecov: 👍

drazisil-codecov (approved):

PR #1069, opened 2023-05-26
feature: support basic authentication when using a proxy

### LINKED ISSUE #1067: Feature Request: Support basic authentication when using a proxy
**Is your feature request related to a problem? Please describe.**

According to the [documentation](https://docs.codecov.com/docs/codecov-uploader#uploader-command-line-arguments), we should use the -U / --upstream args to enable proxy support. However, after trying this and examining the implementation details, it seems that this switch doesn't work for a proxy that requires basic authentication, such as  --upstream=http://alice:pass@proxy.com/ .Could you please let me know if I'm mistaken and if there's a way to use codecov_uploader with a proxy that requires authentication?

also, as we see [from the code](https://github.com/codecov/uploader/blob/3be850357ecc23b661f309c8bc1c71cff3eae172/src/helpers/web.ts#L172), `ProxyAgent` from `undici` is used for the call, and it seems that "[auth mode](https://github.com/nodejs/undici/blob/main/docs/api/ProxyAgent.md#example---basic-proxy-request-with-authentication)" is not supported for [string argument](https://github.com/nodejs/undici/blob/c5f3bdda4214b172e578b6adb933b1b4a7858195/lib/proxy-agent.js#L50-L52) passed by uploader.

This works:
       `"codecov_uploader -r ${repo} -C ${revision} -u ${server} / -f /coverage.xml -X network -F ${test_name}"`

But this doesn't work:
`     "codecov_uploader --upstream http://user:pass@private_proxy.com/  -r ${repo}  -C ${revision} -u ${server} -f coverage.xml -X network -F ${test_name}"`

**Describe the solution you'd like**

A way to pass proxy credential to the uploader

**Additional context**

Opened on behalf of a customer

drazisil-codecov:
closes #1067


src/helpers/proxy.ts
```diff
@@ -0,0 +1,6 @@
+import { ProxyAgent } from 'undici';
+import { UploaderArgs, UploaderEnvs } from '../types';
+
+export function addProxyIfNeeded(envs: UploaderEnvs, args: UploaderArgs): ProxyAgent | undefined {
```
thomasrockhu-codecov, line 4: I'm not sure this needs to be in it's own file. It's only being used in `web`

src/helpers/web.ts
```diff
@@ -162,41 +168,62 @@ export function generateRequestHeadersPOST(
...
-      },
+      headers,
       method: 'POST',
       origin: postURL,
       path: `${url.pathname}${url.search}`,
     },
   }
 }
 
+export function generateHeaders(envs: UploaderEnvs, initialHeaders: IncomingHttpHeaders): IncomingHttpHeaders {
```
thomasrockhu-codecov, line 197: I'm not convinced this needs to be it's own function. It feels particularly dirty since the function is called `generateHeaders`, but we're giving it headers already. I think we should deal with all header generation here.

thomasrockhu-codecov (changes requested):
Thanks for all the cleanup changes! I know I'm guilty of it, but for future PRs, maybe put them in a separate PR. Makes it a lot easier to review.

src/helpers/web.ts
drazisil-codecov: I didn't want a code branch here, since POST and PUT use different headers.

src/helpers/proxy.ts
drazisil-codecov: I thought it was easier to locate it if it needed to be changed, but I can put it back in web if you would like.

src/helpers/web.ts
thomasrockhu-codecov: could we just name it something like `addProxyHeaders` then?

src/helpers/proxy.ts
thomasrockhu-codecov: we could put the `generateHeaders` function here then if we want proxy stuff to be in one file.

thomasrockhu-codecov (approved):
LGTM barring CI

PR #1084, opened 2023-06-19
fix: use token param for upstream proxy auth

SUNx2YCH:
Hello from Lyft!

Unfortunately, [new feature for upstream proxy auth](https://github.com/codecov/uploader/pull/1069) is not work as expected (at least with our proxy), so I changed `ProxyAgent` init according to [undici docs](https://github.com/nodejs/undici/blob/main/docs/api/ProxyAgent.md#example---basic-proxy-request-with-authentication) (additional auth headers were replaced with `token` param).
Seems work fine in our environment.
Note: JS/TS is not my day-to-day language, so more idiomatic options are very welcomed (especially for `getAgentProperty` func with reflection hacks).


SUNx2YCH:
PTAL @drazisil-codecov @thomasrockhu-codecov

SUNx2YCH:
@drazisil-codecov yeah, it's ready
I marked it as draft, because wasn't sure about code style and "idiomaticity"

drazisil-codecov (approved):
Look great, thank you!

PR #1092, opened 2023-07-05
Do not upload test reports to Codecov 

### LINKED ISSUE #1091: Do not upload test reports to Codecov
Current the uploader is locating and uploading test reports that are intentionall broken.

This issue is to prevent that.

drazisil-codecov:
Do not upload test reports to Codecov 
Fixes #1091

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -435,7 +435,7 @@ jobs:
           at: .
       - run:
           name: Remove test coverage files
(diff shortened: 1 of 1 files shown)


thomasrockhu-codecov (approved):
This is fine, would prefer if we had an `if CI` sort of flag, but shouldn't hold up on getting this in

Issue #537, opened 2021-12-09
The uploader should retry on `ECONNRESET` but it doesn't


webknjaz:
**Describe the bug**

A connection reset made a GHA job fail in the codecov action.

**To Reproduce**

https://github.com/cherrypy/cheroot/runs/4465611126?check_suite_focus=true#step:18:16

**Expected behavior**

$sbj

**Screenshots**

https://github.com/cherrypy/cheroot/runs/4465611126?check_suite_focus=true#step:18:16

**Additional context**

Log:
```console
Run codecov/codecov-action@v2
  with:
    files: .test-results/pytest/cov.xml
    flags: GHA, macOS, 3.9, cheroot-8.5.3.dev166-py2.py3-none-any.whl
  env:
    PIP_DISABLE_PIP_VERSION_CHECK: 1
    PIP_NO_PYTHON_VERSION_WARNING: 1
    PIP_NO_WARN_SCRIPT_LOCATION: 1
    PYTEST_ADDOPTS: 
    PY_COLORS: 1
    TOXENV: python
    TOX_PARALLEL_NO_SPINNER: 1
    pythonLocation: /Users/runner/hostedtoolcache/Python/3.9.9/x64
==> macos OS detected
Error: read ECONNRESET
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:201:27)
Post Grab the source from Git
```

webknjaz:
@drazisil-codecov could you apply the `bug` label to this issue as it deserves.

webknjaz:
Got another case of this with a similar but slightly different log @ https://github.com/cherrypy/cheroot/runs/4574188269?check_suite_focus=true#step:18:14
```console
Run codecov/codecov-action@v2
  with:
    files: .test-results/pytest/cov.xml
    flags: GHA, macOS, 2.7, cheroot-8.5.3.dev189+g5dc8b1f7-py2.py3-none-any.whl
  env:
    PIP_DISABLE_PIP_VERSION_CHECK: 1
    PIP_NO_PYTHON_VERSION_WARNING: 1
    PIP_NO_WARN_SCRIPT_LOCATION: 1
    PYTEST_ADDOPTS: 
    PY_COLORS: 1
    TOXENV: python
    TOX_PARALLEL_NO_SPINNER: 1
    pythonLocation: /Users/runner/hostedtoolcache/Python/2.7.18/x64
==> macos OS detected
Error: Client network socket disconnected before secure TLS connection was established
[...]

PR #1098, opened 2023-07-26
fix: retry on ECONNRESET

giovanni-guidini:
Retry the requests POST to codecov and PUT to storage in case of ECONNRESET. Actually it retries in the following cases:
* `errors.UndiciError` with code `'ECONNRESET'`;
* `errors.ConnectTimeoutError`;
* `errors.SockerError`.

That is because I didn't actually know what error ECONNRESET is. I used `UndiciError` as the error base class because I noticed we use the `request` from `undici`, so I figured that if an error happens then it would probably be wrapped by the lib.

Fixes codecov/uploader#537


src/helpers/web.ts
```diff
@@ -15,9 +12,15 @@ import {
...
   UploaderInputs,
 } from '../types'
-import { info } from './logger'
+import { UploadLogger, info } from './logger'
 import { addProxyIfNeeded } from './proxy'
 
+import { promisify } from 'util'
+import { setTimeout } from 'timers'
+
+const maxRetries = 3
```
thomasrockhu-codecov, line 21: I think we can bump this up, or at least increase `baseBackoffDelayMs`. We are waiting only a couple seconds in total

src/helpers/web.ts
```diff
@@ -51,12 +54,35 @@ export function getPackage(source: string): string {
...
+  } catch (error: unknown) {
+    if (
+      ((error instanceof errors.UndiciError && error.code == 'ECONNRESET') ||
+        error instanceof errors.ConnectTimeoutError ||
+        error instanceof errors.SocketError) &&
+      retryCount < maxRetries
+    ) {
+      const backoffDelay = baseBackoffDelayMs * 2 ** retryCount
+      await promisify(setTimeout)(backoffDelay)
+      UploadLogger.verbose('Request to Codecov failed. Retrying...')
```
thomasrockhu-codecov, line 74: can we also log the error?

src/helpers/web.ts
```diff
@@ -69,8 +95,11 @@ export async function uploadToCodecovPUT(
   if (requestHeaders.agent) {
     setGlobalDispatcher(requestHeaders.agent)
   }
-  dns.setDefaultResultOrder('ipv4first');
-  const response = await request(requestHeaders.url.origin, requestHeaders.options)
+  dns.setDefaultResultOrder('ipv4first')
+  const response = await requestWithRetry(
```
thomasrockhu-codecov, line 99: nice

thomasrockhu-codecov (changes requested):

src/helpers/web.ts
giovanni-guidini: I've updated it to retry after 1, 2, 4 and 8 seconds (so total wait of 15s + the request times)

Do you think that's enough?

thomasrockhu-codecov: let's try it for now and see what happens!

thomasrockhu-codecov (approved):
Pending var change

src/helpers/web.ts
```diff
@@ -51,12 +53,36 @@ export function getPackage(source: string): string {
...
+  options: Dispatcher.RequestOptions,
+  retryCount = 0,
+): Promise<Dispatcher.ResponseData> {
+  try {
+    const response = await request(url, options)
+    return response
+  } catch (error: unknown) {
+    if (
+      ((error instanceof errors.UndiciError && error.code == 'ECONNRESET') ||
+        error instanceof errors.ConnectTimeoutError ||
```
webknjaz, line 67: Looks like this doesn't catch `ETIMEDOUT`.

webknjaz:
@giovanni-guidini apparently, this doesn't cover all the cases. See: https://github.com/codecov/uploader/issues/1180.

Issue #544, opened 2021-12-16
Glob path support for Windows drive letters


drazisil-codecov:
`[2021-10-27T15:59:01.743Z] ['verbose'] Possible coverage patterns, after globbing **/C:\Users\circleci\project\coverage\cobertura-coverage.xml`

Looks like we need to strip the glob prefix if we detect we have a drive letter.

_Originally posted by @drazisil-codecov in https://github.com/codecov/uploader/issues/464#issuecomment-953379769_

drazisil-codecov:
@mitchell-codecov do you remember if we fixed this?

mitchell-codecov:
Not to my knowledge. Something we did learn was that the library we use for globbing has Windows-oriented capabilities which may be of interest.

trent-codecov:
@joseph-sentry Please take a look here if we still need to do this and work with @dana-yaish for a solution.

PR #1099, opened 2023-07-26
fix: handle windows drive letter at start of path

joseph-sentry:
Fixes: https://github.com/codecov/uploader/issues/544

### DIFF
--- a/src/helpers/files.ts
+++ b/src/helpers/files.ts
@@ -220,7 +220,7 @@ export async function getCoverageFiles(
 }
 
 export function fetchGitRoot(): string {
-  const currentWorkingDirectory = process.cwd() 
+  const currentWorkingDirectory = process.cwd()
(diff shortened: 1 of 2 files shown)


src/helpers/files.ts
```diff
@@ -320,7 +320,8 @@ export function getFilePath(projectRoot: string, filePath: string): string {
     filePath.startsWith('./') ||
     filePath.startsWith('/') ||
     filePath.startsWith('.\\') ||
-    filePath.startsWith('.\\')
+    filePath.startsWith('.\\') ||
+    /^[A-Z]:\\\S*/.test(filePath)
```
thomasrockhu-codecov, line 324: for sanity, just add a comment here why we are doing this

thomasrockhu-codecov (approved):
small nit

PR #1250, opened 2023-10-20
feat: add user agent to requests to the codecov api

matt-codecov:
add `User-Agent: codecov-uploader/{version}` header to outgoing requests to the codecov api so we can track API usage compared to gazebo and `codecov-cli` or by uploader version

it appears we send a `package` param to a legacy upload endpoint with this information already which can tell us what proportion of that endpoint's traffic is from this uploader, but setting a user agent across the board tells us for all endpoints, or what proportion of overall API traffic is from this uploader

### DIFF
--- a/src/helpers/web.ts
+++ b/src/helpers/web.ts
@@ -19,7 +19,7 @@ import { sleep } from './util'
 
 const maxRetries = 4
 const baseBackoffDelayMs = 1000 // Adjust this value based on your needs.
-export const userAgent: string = 'codecov-uploader/' + version
+export const userAgent: string = `codecov-uploader/${version}`
 
 /**
  *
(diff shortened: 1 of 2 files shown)


src/helpers/web.ts
```diff
@@ -19,6 +19,7 @@ import { sleep } from './util'
 
 const maxRetries = 4
 const baseBackoffDelayMs = 1000 // Adjust this value based on your needs.
+export const userAgent: string = 'codecov-uploader/' + version
```
thomasrockhu-codecov, line 22:
```suggestion
export const userAgent: string = `codecov-uploader/${version}`
```

thomasrockhu-codecov (approved):
Small nit

PR #1312, opened 2023-11-17
chore(ci): first pass at build-linux

thomasrockhu-codecov:


drazisil-codecov (changes requested):
We appear to be missing the review job, which is where we can check all the logs to ensure there are no errors.

drazisil-codecov (approved):
I would _really_ like the review job to be prioritized, but ok.

PR #1362, opened 2023-12-14
fix: update 'type' file import

zhangtao25:
The import of the type file in the proxy.ts file is wrong.


zhangtao25:
> I don't believe it is. Can you please explain?

possibly an error caused by the ai complement.

PR #1380, opened 2024-01-08
feat: support longer flag names that can contain special chars

joseph-sentry:


matt-codecov (approved):
maybe add a test case with a 1025-char flag name

thomasrockhu-codecov (approved):

Issue #1773, opened 2024-11-07
`test and deploy` workflow is broken


thomasrockhu-codecov:
https://github.com/codecov/uploader/actions/workflows/test-and-deploy.yml

PR #1775, opened 2024-11-11
fix: test-and-deploy mac builds

thomasrockhu-codecov:
fixes https://github.com/codecov/uploader/issues/1773


trent-codecov (approved):
Do we need to add this to our releaser?

